PR #92, opened 2026-06-13
Optional closed-beta referral gate at signup

jlamoreaux:
## What

Adds an **optional** closed-beta gate: when enabled, **no one can create an account without a valid referral code**. It's the app-side enforcement for the referral service in `jlamoreaux/stratum-landing#7`.

**Fully opt-in / no-op by default.** The gate is active only when `BETA_GATE="1"` **and** `REFERRAL_SERVICE_URL` are set, so OSS self-hosters are unaffected. Existing auth tests pass unchanged.

### Behavior when the gate is ON
- **Magic-link signup** — an invite code is required; validated before the email is sent and re-validated before `createUser`. After creation, the user's **5** codes are minted via the service and emailed (`getInviteCodesEmail`).
- **GitHub / Google OAuth** — login-only during beta. A first-time OAuth user with no existing account is redirected to `/auth/signup?error=invite_required`, so OAuth can't bypass the gate.
- The landing **waitlist stays open** (no code needed to join); the gate is on account creation, not the waitlist.

### Files
- `src/beta/gate.ts` — `betaGateEnabled` / `validateInviteCode` / `admitUser`. Fails closed on service errors; `admitUser` never throws into the signup path.
- `src/routes/email-auth.tsx`, `src/routes/auth.ts`, `src/routes/signup.tsx` — wiring at the three account-creation paths + a conditional invite field (prefilled from `?ref=`).
- `src/email/templates.ts` — `getInviteCodesEmail`.
- `src/types.ts`, `wrangler.toml` — new optional env vars.
- `tests/beta-gate.test.ts` — 10 unit tests (enable/disable, fail-closed, bearer auth).

## How to enable (Stratum Cloud)

```bash
cd stratum
```

1. In `wrangler.toml` `[vars]`, uncomment:
   ```toml
   BETA_GATE = "1"
   REFERRAL_SERVICE_URL = "https://usestratum.dev"
   ```
2. Set the shared secret — **must equal the value set on stratum-landing**:
   ```bash
   wrangler secret put REFERRAL_SERVICE_SECRET
   ```
3. `wrangler deploy`.

**Local dev:** the two `[vars]` apply automatically; put the secret in `stratum/.dev.vars`:
```
REFERRAL_SERVICE_SECRET=dev-secret-value
```
Point `REFERRAL_SERVICE_URL` at your local landing (`wrangler dev`) to exercise the flow. To disable the gate, comment out `BETA_GATE` (or set it to anything other than `"1"`) and redeploy.

## Depends on
`jlamoreaux/stratum-landing#7` (the referral service + `REFERRAL_SERVICE_SECRET` it shares). Until seed codes are minted there, account creation is fully closed — expected for a closed beta.

## Summary by CodeRabbit

* **New Features**
  * Closed-beta invite-code gating for sign-ups (includes OAuth and magic-link flows)
  * Users receive shareable invite codes emailed to them after signup
  * Signup form shows an invite-code field when the beta gate is active

* **Tests**
  * Added tests covering gate logic, invite validation, and admit/redeem behaviors

* **Documentation**
  * Added configuration guidance for enabling the closed-beta gate in deployment config

### DIFF
--- a/src/beta/gate.ts
+++ b/src/beta/gate.ts
@@ -1,0 +1,112 @@
+/**
+ * Closed-beta gate (Stratum Cloud only).
+ *
+ * This is the ONLY core hook into the referral/beta program. The program itself
+ * lives in the cloud/landing layer; core just calls it. When the gate env vars
+ * are unset (the default for OSS self-hosters) every function here is inert and
+ * account creation is unchanged.
+ *
+ * - betaGateEnabled: is the gate switched on and pointed at a service?
+ * - validateInviteCode: is this code redeemable? (pre-createUser check)
+ * - admitUser: record the redemption + mint the user's 5 codes (post-createUser)
+ */
+import type { Env } from "../types";
+import type { Logger } from "../utils/logger";
+
+export interface InviteValidation {
+  valid: boolean;
+  referrerUserId: string | null;
+}
+
+export interface AdmitResult {
+  codes: string[];
+  referrerUserId: string | null;
+}
+
+/** True only when the gate is explicitly enabled AND a service URL is configured. */
+export function betaGateEnabled(env: Env): boolean {
+  return env.BETA_GATE === "1" && !!env.REFERRAL_SERVICE_URL;
+}
+
+function serviceUrl(env: Env, path: string): string {
+  return `${(env.REFERRAL_SERVICE_URL ?? "").replace(/\/$/, "")}${path}`;
+}
+
+/**
+ * Check whether an invite code can currently be redeemed. Fails closed
+ * ({ valid: false }) on any network/parse error so a service outage cannot
+ * silently let ungated users through the beta wall.
+ */
+export async function validateInviteCode(
+  env: Env,
+  code: string,
+  logger: Logger,
+): Promise<InviteValidation> {
+  const trimmed = (code ?? "").trim().toUpperCase();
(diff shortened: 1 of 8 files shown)


jlamoreaux:
Thanks @coderabbitai — addressed in the latest commit:

- **Timeouts on referral-service fetches** ✅ — both `validateInviteCode` and `admitUser` now use `AbortSignal.timeout(5000)`, so a hung service can't stall signup (validate runs inline in the request).
- **Verified-email gate check (auth.ts)** ✅ — the GitHub beta-gate existence check now matches an existing account only by a *verified* email, never the unverified `primaryEmail` fallback (which could be attacker-controlled).
- **Test for `admitUser` non-OK branch** ✅ — added a 401 case asserting `{ codes: [], referrerUserId: null }`.

**Intentionally not applied:** requiring `REFERRAL_SERVICE_SECRET` in `betaGateEnabled`. Adding it would make the gate **fail-open** — if an operator sets `BETA_GATE=1` + `REFERRAL_SERVICE_URL` but forgets the secret, `betaGateEnabled` would return `false` and signups would open entirely. The current behavior is safer: the gate stays closed (codes are still required via `validateInviteCode`, which needs no secret), and only the post-signup code-minting degrades (logged). A missing secret should never silently disable the wall.

PR #114, opened 2026-06-17
ci: stop opening an issue on every successful production deploy

jlamoreaux:
### Problem
The `deploy-production` job in `ci.yml` had a **"Create deployment notification"** step (`if: success()`) that called `github.rest.issues.create(...)` on every successful production deploy — so each deploy filed a new **"✅ Production Deployment Successful"** issue. That's unwanted noise in the Issues tab.

### Fix
Replace the issue creation with a write to the **GitHub Actions run summary** (`core.summary`). The same deployment record (commit, message, URL, checklist) still shows up — on the workflow run's summary page instead of as an issue.

- The `Notify on failure` step (`if: failure()`) is **unchanged** — real failures still notify.
- This is the only `issues.create` guarded by `success()`; all others across the workflows are `failure()`-gated.

### Verified
- `ci.yml` parses as valid YAML.

### Note
This doesn't retroactively close the issues already created by past deploys. Happy to bulk-close any open `✅ Production Deployment Successful` issues (label `deployment`/`production`) in a follow-up if you'd like.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -200,7 +200,7 @@ jobs:
           npx vitest run tests/smoke/ --reporter=verbose || true
           echo "✓ Smoke tests completed"
 
-      - name: Create deployment notification
+      - name: Record deployment summary
         if: success()
         uses: actions/github-script@v7
         with:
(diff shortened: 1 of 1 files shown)


jkl3848:
lol

Issue #115, opened 2026-06-18
Native git push: smart-HTTP proxy so a Stratum repo can be set as origin


jlamoreaux:
## Summary

Stratum has no native `git push`. You cannot set a Stratum repo as `origin` and push to it with stock git. The only write path is the REST API (`stratum commit`, which ships staged file *contents* as JSON; the Worker turns that into a server-side `commitAndPush`).

This is a recurring adoption objection — existing repos, CI publish steps, and developer muscle memory all assume `origin` is pushable.

## Why it's impossible today

- The real backing remote is a **Cloudflare Artifacts URL** (`*.artifacts.cloudflare.net/git/…`), an implementation detail, not a public interface.
- Auth is a **short-lived Artifacts token minted server-side** (`freshRepoToken` in `src/storage/git-ops.ts`, HTTP Basic with `username: "x"`). No route vends it to clients; a client can't produce one.
- Stratum's own domain hosts **no git endpoint** — there are no `info/refs` / `git-upload-pack` / `git-receive-pack` routes in `src/index.ts`.
- A raw direct push would **bypass the change/eval gate, merge queue, activity, and protections**.

## Proposal

Add a **git smart-HTTP proxy** mounted on the Stratum domain (e.g. `https://app.usestratum.dev/@ns/slug.git`) that:

1. Authenticates with the existing API-key system (key as HTTP Basic password — no new credential type).
2. Authorizes the caller (read for `upload-pack`, write for `receive-pack`).
3. Mints a fresh Artifacts token with `freshRepoToken` and proxies upstream; the token never leaves the Worker.
4. Routes `receive-pack` pushes through the existing change → eval → `MergeQueue` pipeline so a `git push` produces the same artifacts as `stratum commit` + change, and protected refs stay gated.

```
git remote add origin https://x:<stratum_user_…>@app.usestratum.dev/@alice/my-project.git
```

## Known hard parts

- **Streaming**: `upload-pack`/`receive-pack` are half-duplex and Workers can't half-duplex stream (same wall as ADR 004's merge push) — pack bodies must be buffered/chunked within Worker limits.
- Mapping eval rejection onto git's sideband so `git push` fails legibly.
- SSH transport is out of scope (Workers have no raw TCP listener).

## Suggested slicing

Ship read-only `upload-pack` (clone/fetch) first as the smaller, lower-risk slice; follow with `receive-pack` (push) once streaming is proven.

## Design doc

[...]

PR #123, opened 2026-06-18
fix(merge): merge the evaluated commit, not the workspace's live tip (#115)

jlamoreaux:
## Problem (gate soundness / TOCTOU)

The merge gate evaluates a change against the workspace tip, then merges whatever the workspace's `main` points at **at merge time**. A re-push between evaluation and merge lands **unevaluated content** on the project default branch — defeating the gate, for any change source (CLI `stratum commit`, REST, or git push). This is PR B1 of completing #115 (gated `git push`), but it's an independent correctness fix.

## Fix — pin and merge the evaluated commit

- **Migration 024**: `changes.workspace_head_sha` records the commit a change was evaluated against.
- **`getDiffBetweenRepos`** now returns `{ diff, workspaceSha }`, resolving the sha from the **same clone** it diffs — so the pinned sha is provably the evaluated content. (Also removes a redundant second clone that an earlier draft used, which had a cross-clone race — caught in review.)
- **`mergeWorkspaceIntoProject`** gains an optional `workspaceSha`: when set it `readCommit`-verifies the sha is reachable in the freshly-fetched workspace and merges **that** commit (3-way **and** squash). If a re-push rewound past it, the merge **fails closed** rather than merging a different tip. Omitted ⇒ legacy live-tip behavior (NULL on pre-migration changes — backward compatible).
- Threaded through the **production** merge paths: the cold `POST /changes/:id/merge` path and `MergeQueue.merge`.

## Scope / performance

Production runs `REPO_DO_ENABLED=false`, so **every production merge now goes through the pinned path**. The staging-only R2/group-commit fast path (`mergeViaR2`/`fastForwardMerge`/`batchMergeStagedTrees`) still resolves the live tip — it must be pinned **before `REPO_DO_ENABLED=true` ships to prod** (follow-up). That path is **untouched here**, so the throughput benchmark is unchanged:

```
group-commit: 431.1 c/s   (baseline 431.1; target 22.6) — no regression
```

## Tests / review

- Unit: change-create persists the pinned sha (from the diff clone); merge threads `workspaceSha` into `mergeWorkspaceIntoProject`; the no-pin (legacy) path still passes `{strategy}` only.
- Typecheck ✅ · full suite 856 ✅ · Biome ✅ · throughput benchmark ✅.
- A focused security review confirmed fail-closed behavior, squash pinning, reachability, production routing, backward-compat, and that the cross-clone capture race is closed.

⚠️ Follow-up (blocking for R2-in-prod): pin the R2/group-commit fast path to the evaluated sha/tree.

## Summary by CodeRabbit

* **Bug Fixes**
  * Workspace commit SHA is now captured during change evaluation and pinned to each change record, ensuring merges execute against the exact evaluated workspace state rather than the workspace's current live tip.

* **Tests**
  * Added test coverage to verify merged changes use the pinned evaluated workspace commit.

### DIFF
--- a/migrations/024_change_workspace_head_sha.sql
+++ b/migrations/024_change_workspace_head_sha.sql
@@ -1,0 +1,5 @@
+-- The exact workspace commit sha that a change was evaluated against. The merge
+-- gate must merge *this* commit, not the workspace branch's current tip — pinning
+-- it closes a TOCTOU where a re-push between evaluation and merge could land
+-- unevaluated content on the project's default branch (ADR 005 / #115).
+ALTER TABLE changes ADD COLUMN workspace_head_sha TEXT;
--- a/src/queue/merge-queue.ts
+++ b/src/queue/merge-queue.ts
@@ -64,7 +64,12 @@ export class MergeQueue extends DurableObject<Env> {
         workspace.remote,
         workspaceToken.data,
         log,
-        { strategy: "merge", timer },
+        // Merge the exact evaluated commit, not the workspace's live tip (#115).
+        {
+          strategy: "merge",
+          timer,
+          ...(change.workspaceHeadSha ? { workspaceSha: change.workspaceHeadSha } : {}),
+        },
       );
 
       if (!commitResult.success) {
--- a/src/routes/changes.ts
+++ b/src/routes/changes.ts
@@ -279,7 +279,9 @@ app.post("/projects/:name/changes", async (c) => {
     logger.error("Failed to get diff between repos", diffResult.error);
     return badRequest(diffResult.error.message);
   }
-  const diff = diffResult.data;
+  // Pin the exact commit the diff was computed from, so the merge consumes this
+  // sha rather than a later re-pushed tip (gate soundness — #115).
+  const { diff, workspaceSha: workspaceHeadSha } = diffResult.data;
 
   const evaluators: Array<{ type: string; evaluator: Evaluator }> = [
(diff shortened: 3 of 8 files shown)


jlamoreaux:
## Real throughput measured on staging (not just the local model)

The PR body cited "431 c/s" — that's the **local in-process model** (`tests/group-commit.test.ts`, simulated ~50 ms write, no real network/D1/R2). I ran the **real harness** (`scripts/bench-commit-throughput.ts`) against staging to back the claim with an actual number.

**Group-commit / batch-merge path** (the ADR 004 target path):

| N | wall c/s | server-side c/s |
|---|---|---|
| 25 | 9.15 | 11.85 |
| 50 | 19.84 | **33.18** |
| 64 | 17.32 | 23.01 |

**Full per-change commit→merge** (non-batched): N=1 → 0.49, N=5 → 1.05, N=25 → 2.24 c/s.

Takeaways:
- **Peak server-side ≈ 33 c/s (N=50), clearing the 26 c/s target** on real infra via batching. Wall-clock (~20 c/s) includes client→edge round-trips + per-change setup.
- The local "431" is a ~13× optimistic model ceiling, not reality.
- **This PR is throughput-neutral**: it doesn't touch the group-commit/R2 path, so these numbers are unchanged vs `main` (the pinning only affects the cold per-change merge, which is eval/clone-bound at ~2 c/s regardless).

Method note: measured on the B1 branch deployed to staging (required applying migration 024 to staging D1); staging has since been restored to `main`.

jlamoreaux:
@coderabbit review

PR #125, opened 2026-07-09
feat(deletion): GDPR-grade project & account erasure — cascade, verifier, durable job runner, enforcement

jlamoreaux:
Durable, GDPR-grade deletion for projects and accounts. This PR consolidates what was originally a 4-PR stack (#126, #127, #128 merged down into this branch) plus a round of review hardening, so it lands as one coherent change to `main`.

## What it does

**Project deletion** — cascade-deletes a project and everything scoped to it (workspaces, fork repos, changes, webhooks, provenance, the Artifacts repo, KV identity), then verifies the deletion actually completed and reports anything left behind as residuals.

**Account erasure** — GDPR-grade self-service erasure: sets `deleting_at` (which immediately revokes the caller's credentials via the auth middleware), enqueues an account cascade, and anonymizes cross-project contributions against a shared tombstone rather than destroying other tenants' history.

Both run as a **D1-backed durable job** driven by the `*/5` sweep — never a best-effort synchronous request that can half-finish. Each job carries the full captured target JSON (so a re-drive after a partial cascade still knows about FK children whose parents are already gone), plus a checkpoint / heartbeat / lease so two drivers can't run the same cascade.

## Architecture

- **Data model (migration 025):** dual-write `project_id` (a stable UUID) on project-scoped D1 tables, so deletion targets a UUID instead of a mutable name — legacy rows with only a bare `project` name are handled explicitly under name collisions.
- **Durable jobs (migration 026):** `deletion_jobs` state machine (`pending → running → verifying → completed | incomplete`) with lease-based single-driver execution.
- **Atomic dedup (migration 027):** a partial unique index (one active job per `kind + target_id`) so concurrent delete requests can't enqueue duplicate cascades.
- **Enforcement:** `isTargetDeleting` gates merges, imports, and syncs against a repo whose project or owner is being deleted.
- **Surfaces:** `DELETE /api/projects/:ns/:slug` and `DELETE /api/users/me` (confirm-token guarded), UI Danger Zones, CLI commands, and OpenAPI docs.

## Review hardening (CodeRabbit, addressed in this branch)

- Project-scoped deletion now blocks in-flight writes across all four enforcement sites (repo-do, merge-queue, import-queue, sync), not just account deletion.
- Route-level find-then-insert TOCTOU replaced with the partial unique index + conflict handling in `createDeletionJob`.
- `findActiveJobForTarget` escapes LIKE wildcards (ids contain `_`) with an `ESCAPE` clause.
- `RepoDO.purge` runs its wipe inside `blockConcurrencyWhile` (a deeper in-flight-resume refinement is tracked in STR-14).
- `projectId` backfilled on the legacy `change.merged` / `change.rejected` / evaluate-cost paths.
- Route audit emits `deletion.requested`; the runner keeps `deletion.started` (no duplicate).
- OpenAPI `401` on `DELETE /api/users/me`; `REPO_DO_ENABLED` added to the worker types.

## Tests

Full suite green — **920 passed**, typecheck + lint clean. Coverage includes the cascade → verify → residuals flow, account cascade, deleting enforcement, lease/sweep, atomic dedup (storage + route), and project-scoped enforcement.

## Migrations

Deploy `025` → `026` → `027` in order.


jlamoreaux:
Thanks @coderabbitai — addressed in the latest commit:

- **CRITICAL (merge-batch provenance INSERT > 100 params):** fixed. `INSERT_CHUNK` 12 → **11** (11×9 = 99 ≤ 100) and corrected the "8 binds/row" comment to 9. Confirmed the UPDATE chunk stays within cap.
- **Backfill project_id on the classic merge path (changes.ts) and review events (reviews.ts):** done — now `change.projectId ?? project.id` (project is in scope), matching the merge-queue/repo-do pattern so legacy changes don't persist NULL.
- **github-bridge dual-write:** skipped intentionally. `upsertChangeFromGitHubPR` has no callers (dead code) and already bound the UUID to `project` on `main` (pre-existing, not introduced here); the legacy name isn't in scope and `project_id` correctly holds the UUID. Happy to fix if/when that path goes live.

The fix is rebased through the stacked PRs (#126 → #127 → #128). Typecheck, full suite, and Biome all green.

jlamoreaux:
@coderabbitai review

jlamoreaux:
@coderabbitai full review

jlamoreaux:
@coderabbitai review

PR #133, opened 2026-07-19
Security hardening: close audit findings (auth, merge gate, provenance)

jlamoreaux:
Closes the exploitable and integrity gaps from the 2026-07-18 principal-eng dogfood audit. Scoped, tested, and reviewed.

## Findings addressed

| ID | Severity | Fix |
|----|----------|-----|
| SEC-1 | **Critical** | Remove the unauthenticated `POST /api/projects/:name/sync` route (anyone could repoint a project's `githubUrl` and trigger a destructive re-import). Namespaced authenticated route unaffected. |
| SEC-2 | **High** | Bind the merge gate to the evaluated revision. `getDiffBetweenRepos` returns the workspace tip oid from the *same* clone the diff used; stored as `changes.evaluated_sha` at eval time. Every merge backend re-checks the tip in the shared pre-branch block and returns `409 STALE_WORKSPACE` if it moved; **fails closed** (`409 WORKSPACE_UNVERIFIABLE`) when the tip can't be resolved. |
| SEC-3 | **High** | Force merge is **deny-by-default** (opt in via `merge.allowForce: true`) in single + batch paths. |
| SEC-4 | Medium | Scope by-id reads: agent `GET /:id` owner-only; org `GET /:slug` + `/:slug/teams` member-only. 404 (no existence leak) to non-owners/members and anonymous. |
| SEC-5 | Medium | Constant-time admin-key compare; gate `/dev-login` on `DEV_LOGIN_ENABLED` (local-only) as well as the localhost check. |
| SEC-6 | Medium | Private-host filter + `redirect: "manual"` on the webhook evaluator and delivery fetches (SSRF). |
| SEC-7 | Low | Global security headers (nosniff, X-Frame-Options, Referrer-Policy, HSTS, non-`script-src` CSP that won't break inline UI handlers); git smart-HTTP exempt. |
| Provenance | Integrity | Snapshot `agent_model` + `agent_prompt_hash` on the change at creation; record `model` + `prompt_hash` on the provenance row across **all five** write sites (incl. the raw batch INSERT), so the audit trail matches the product claim. |

## Approach notes

- **SEC-2** is the subtle one. It captures the evaluated sha from the same workspace clone the diff was computed against (avoiding a TOCTOU that a naive second `getCommitLog` would open), and the check sits before the RepoDO/MergeQueue/cold branch so it covers every backend uniformly — the R2/DO paths merge a *staged tree* without a live clone, which is exactly why a silently-skipped check would be a real bypass. Hence fail-closed.
- Migrations `024` (changes: `evaluated_sha`, `agent_model`, `agent_prompt_hash`) and `025` (provenance: `model`, `prompt_hash`) are additive; historical rows keep NULLs.

## Deferred (out of scope, tracked)

Rate-limiter atomicity (needs a DO), secrets-at-rest crypto rotation (static PBKDF2 salt / cleartext webhook secrets), magic-link atomic single-use, CSP `script-src` tightening (needs an inline-handler refactor).

## Validation

- Lint (Biome) clean · typecheck clean · **889 tests pass (+34 new)**
- Both migrations apply cleanly (`wrangler d1 migrations apply --local`)
- A parallel bugs/conventions/edge-cases review ran on the diff; the fail-open it surfaced in the merge gate was made fail-closed (both single and batch paths).

## Summary by CodeRabbit

* **Security**
  * Strengthened access controls for agents, organizations, teams, workspaces, imports, and synchronization.
  * Added SSRF protection, security headers, safer token handling, and stricter webhook redirect behavior.
* **Bug Fixes**
  * Prevented merges against stale or changed workspaces and blocked malformed policy configurations.
  * Limited workspace commit sizes and improved event retry handling.
* **New Features**
  * Added provenance details for AI-assisted changes, including model and prompt metadata.
  * Added local-only development login protection.
* **Breaking Changes**
  * Force merges are now disabled by default; explicitly enable them in project policy.


jlamoreaux:
Thanks for the review — addressed in `7de6946`.

**🔴 Provenance batch INSERT exceeds D1's 100-param limit** — Fixed. Real regression I introduced: adding `model` + `prompt_hash` made each row bind 10 params, so `INSERT_CHUNK = 12` → 120. Now derived from the bind count (`Math.floor(100 / 10) = 10` rows/statement).

**🔴 Bind every merge input to the evaluated revision (TOCTOU)** — Fixed properly. The pre-merge tip check left a window before the staged-tree read. I now capture the evaluated tip's **tree oid** from the same clone the diff used (`getDiffBetweenRepos` → `workspaceTreeOid`), persist it as `changes.evaluated_tree_oid` (migration `026`), and content-address what each backend lands: `mergeViaR2` rejects and the batch path skips when `staged.treeOid != evaluated_tree_oid`. This is atomic with the merge and network-free, so unlike the tip check it holds even under `force`; legacy changes with no `evaluated_tree_oid` skip it. Added a batch mismatch test.

**🟠 Security headers not applied to thrown 500s** — Fixed. Headers now register before `next()` so they survive the error boundary; git smart-HTTP stays exempt.

**🟠 Fail change creation when the agent snapshot can't be loaded** — Partially addressed. I now **log** the `getAgent` failure (the "never silently swallow" rule), but kept it **best-effort** rather than failing the whole change: provenance metadata shouldn't block the core operation, and a missing model/prompt-hash snapshot degrades gracefully (the fields are nullable). If you'd prefer hard-fail, happy to change it.

**🔴 dev-login localhost check** — Skipped (already present). The condition below the flag check (`index.ts:64-67`) enforces the localhost host, and `dev-login-gate.test.ts` asserts a non-localhost host returns 403 even when the flag is on. CodeRabbit only saw the flag check in the diff.

Full suite green: 890 tests (+ the new mismatch test), lint + typecheck + coverage clean.

jlamoreaux:
Second-round review — CodeRabbit re-posted the same four findings on the fix commit. Verified each against current code:

**🔴 dev-login localhost check** — Now folded into a **single condition** (`46e663c`, `index.ts:62-64`): `DEV_LOGIN_ENABLED !== "true" || host not localhost → 403`. The localhost gate was already enforced (previously as a second `if`); this removes the two-statement form the bot kept flagging. Covered by `dev-login-gate.test.ts` (non-localhost + flag-on → 403).

**🟠 Move `await next()` to the end of security-headers** — Already done in the previous commit. `security-headers.ts:37` has `await next()` as the last statement, with all headers registered before it. This is exactly the suggested shape; the re-flag is stale.

**🔴 Bind every merge input to the evaluated revision (TOCTOU)** — Addressed via content-addressing in the previous commit, which the static rule doesn't trace across files. The check is now *inside* each backend, immediately before the merge: `mergeViaR2` (`repo-do.ts`) rejects when `staged.treeOid !== change.evaluatedTreeOid`, and the batch path skips the same way — both comparing the exact tree about to land against the evaluated tree oid (captured from the same clone as the diff, stored as `changes.evaluated_tree_oid`). That closes the window between the pre-merge tip check and the staged-tree read.

**🟠 Hard-fail change creation on agent-snapshot load failure** — Holding as best-effort **with logging** (`changes.ts`, `logger.warn` on `getAgent` failure). The "never silently swallow — log it" rule is satisfied by the log; failing an entire change over an unreadable *provenance metadata* field (nullable, recorded at merge) would be disproportionate, since the change is a real work product. Open to hard-failing if a maintainer prefers — it's a one-line change.

All green: 890 tests, lint, typecheck, coverage.

jlamoreaux:
Third round (`a1a4cb7`):

**🔴 TOCTOU — now closed on *all* paths.** The staged R2/DO paths were already content-addressed on tree oid. This commit closes the last window on the **cold/squash fallback**: `mergeWorkspaceIntoProject` takes an `expectedWorkspaceSha` and verifies the freshly-fetched tip equals the evaluated sha *immediately before* merging, returning `409 STALE_WORKSPACE` otherwise. So every backend now validates the exact input it's about to merge against what was evaluated — the route's tip check is now just an early-exit optimization, not the guarantee. Applies even under force (uses the already-fetched tip, no extra cost). Added cold-path pin + 409-mapping tests.

**🟠 security-headers `await next()`** — Restructured to a single `await next()` with no early-return branch (headers still register before it; git paths still exempt). Functionally identical to before, but removes the branch the static rule kept mis-reading.

**🟠 Agent-snapshot hard-fail** — Still holding as best-effort-with-logging (see prior comment); it's a deliberate call that a real change shouldn't fail over an unreadable, nullable, merge-time provenance field. Happy to flip on a maintainer's word.

892 tests, lint, typecheck, coverage all green.

jlamoreaux:
## Second-pass audit fixes (8 commits since the review round)

A follow-up principal-eng audit found issues **more severe than the first pass** — including a gap in my own SEC-2 fix — so this branch now also carries:

**Missing write authorization (HIGH):**
- `POST /api/workspaces/:name/commit` and `DELETE /api/workspaces/:name` had **no** project-write check — any authenticated caller who knew a project id (returned by `GET` project) + workspace name could push commits into, or destroy, another tenant's workspace. Both now gate on `canWriteProject`; commit bounds the file map; delete targets the remote-derived Artifacts repo.
- Bulk import trusted a caller-supplied `namespace` → now own-namespace-only (no squatting).

**My SEC-2 fix was incomplete in production:** prod runs the `MergeQueue` backend (`REPO_DO_ENABLED=false`), which I'd left **unpinned** — the TOCTOU was still open there. `MergeQueue.merge`, `RepoDO.advance()` fast-forward, and its cold fallback now all thread `expectedWorkspaceSha`. Every merge backend now validates the merged tip/tree against what was evaluated.

**Other second-tier fixes:** SSRF filter blocks obfuscated IP encodings (int/hex/octal/`[::]`); magic-link tokens are single-use atomically (moved to D1, hashed at rest); malformed policy files fail the merge gate closed instead of silently downgrading; event retries resume past completed handlers instead of duplicating analytics/webhooks. Plus the CLEANUP items (onError full header set, D1 chunk headroom, CHANGELOG breaking-change note).

Migrations `024`–`028` apply cleanly in sequence. **910 tests** (up from the 855 baseline), lint + typecheck + coverage green.

Not attempted here (architecture-scale, flagged for the team): durability/backup, KV→D1 identity migration, merge-backend consolidation, per-project agent-token scoping.

jlamoreaux:
@coderabbitai review

PR #135, opened 2026-07-20
fix(audit): tranche 1 — safe high-value remediation

jlamoreaux:
First tranche of the principal-eng re-audit remediation (`/ship` methodology). Scoped to **localized, no-migration, no-data** fixes so it shares no revert surface with the riskier tranches.

## Findings closed
- **Functional bug (High):** bulk-import created the job and set status `queued` but **never called `queueImportJob`** — bulk-imported repos were never cloned. Now enqueues, mirroring the single-import path.
- **Security (conflict-resolve BFLA, Medium):** `POST /projects/conflicts/:id/resolve` minted a write token and pushed a merge into the project repo with only an *authenticated* check — no `canWriteProject`. Any authed caller holding a (leakable) conflict id could write into another tenant's repo. Now requires project write; collapses to 404 (no existence leak).
- **Security (downgrade, Low):** repo-URL validators accepted `http://`; now `https://` only (host still pinned to the 3 providers).
- **Self-approval (Low), partial:** `countApprovals` gains an `excludeUserId` so a change author's own approval won't satisfy `requiredApprovals`. *Call-site wiring deferred* — the `changes` table has no creating-user column yet (only `agentId`); tracked in `.claude/ship/TASKS.md`.

## Tests
Self-approval exclusion, https-only URLs, conflict-resolve authz denial. **Full suite green (1050 pass), typecheck + lint clean.**

## What's staged for later tranches
Sessions/CSRF + account-enumeration · merge-CAS de-dup · deletion enforcement + re-drive · backup restore/health/all-refs · KV→D1 canonical identity · LIMIT/pagination.

## ⚠️ Gated — need operator input, deliberately NOT in this PR
- **CI `--remote` migration fix:** flipping `--remote` today would apply an unknown accumulated backlog to prod in one shot (staging shares the bug, so isn't a faithful rehearsal). Needs the real remote `d1_migrations` ledger state + reconciliation first.
- **Live `POST /api/admin/restore`:** total-data-loss blast radius behind one shared admin key — shipping `scripts/restore.ts` (dry-run default) instead; a live route needs a break-glass decision.
- **Removing the T5 name-fallback:** gated on verified zero-NULL `project_id` after backfill.

## Summary by CodeRabbit

- **New Features**
  - Repository imports now begin processing automatically after being queued.
  - Conflict resolution now requires project write access.

- **Bug Fixes**
  - Approval counts can exclude approvals from the change author.
  - Repository validation now requires secure HTTPS URLs for supported providers.
  - Unauthorized conflict-resolution requests return a generic not-found response.

- **Tests**
  - Added coverage for approval exclusions, access control, and HTTPS-only repository validation.

### DIFF
--- a/src/merge/protection.ts
+++ b/src/merge/protection.ts
@@ -65,6 +65,9 @@ export async function checkMergeProtection(
 
   const requiredApprovals = merge.requiredApprovals ?? 0;
   if (requiredApprovals > 0) {
+    // NOTE: self-approval exclusion (countApprovals' excludeUserId arg) is wired in a
+    // follow-up — the `changes` table records no creating-user id yet (only agentId),
+    // so excluding the author requires a schema addition. Tracked in TASKS.md.
     const approvalsResult = await countApprovals(db, logger, change.id);
     if (!approvalsResult.success) return err(approvalsResult.error);
     if (approvalsResult.data < requiredApprovals) {
--- a/src/routes/bulk-import.ts
+++ b/src/routes/bulk-import.ts
@@ -192,7 +192,6 @@ export async function processRepoImport(
       return { success: false, error: setResult.error.message, repo: url };
     }
 
-    // Mark as successful (actual import happens in background)
     await updateImportStatus(
       env.DB,
       namespace,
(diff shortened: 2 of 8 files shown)


jlamoreaux:
@coderabbitai review — both prior findings (change_reviews reviewer_id column, and bulk-import enqueue error handling) are already fixed in afbf5f2; the earlier review was on the pre-fix commit 62343dc.

PR #136, opened 2026-07-20
fix(audit): tranche 2 — auth/session hardening (hash sessions, login-CSRF)

jlamoreaux:
Second tranche of the re-audit remediation. Self-contained security fix.

**Finding (Low, but sharp):** session IDs were stored **plaintext** in D1 while every other secret is SHA-256 hashed. Any read-only leak of the `sessions` table (SQLi read, a leaked backup, logs) yielded directly replayable session cookies. Now hashed at rest — the raw id lives only in the cookie.

**Done correctly across all 5 call sites + the list/revoke contract** (the part the PRD critique flagged as easy to get wrong):
- `createSession` stores `hashToken(rawId)`, returns the raw id for the cookie
- `getSession`/`deleteSession`/`refreshSession` hash the raw cookie input
- `getUserSessions` surfaces the stored hash as an opaque handle; list `isCurrent` compares the **hashed** cookie
- revoke-from-list uses a new `deleteSessionByStoredId` (no double-hash)

**Cutover:** existing plaintext sessions stop resolving (forces re-login) — and a restored old backup can no longer reintroduce live session credentials.

Test added; full suite green (1046), typecheck + lint clean.

## Summary by CodeRabbit

* **New Features**
  * Added an explicit confirmation step for email magic-link verification before signing users in.
  * Magic-link verification now uses a secure same-origin form submission.

* **Bug Fixes**
  * Improved session tracking and revocation, including reliably identifying and deleting the current session.
  * Session identifiers are now protected when stored, while existing sign-in sessions continue to work normally.

### DIFF
--- a/src/middleware/csrf.ts
+++ b/src/middleware/csrf.ts
@@ -1,9 +1,56 @@
-import type { MiddlewareHandler } from "hono";
+import type { Context, MiddlewareHandler } from "hono";
 import { isGitHttpPath } from "../routes/git-http";
 import type { Env } from "../types";
-import { createLogger } from "../utils/logger";
+import { type Logger, createLogger } from "../utils/logger";
 
 const STATE_CHANGING_METHODS = new Set(["POST", "PUT", "PATCH", "DELETE"]);
+
+/**
+ * Reject a state-changing request whose Origin/Referer doesn't match the request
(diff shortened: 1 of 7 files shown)


jlamoreaux:
@coderabbitai review

PR #137, opened 2026-07-20
fix(audit): bound KV get fan-out + follow the list cursor (no truncation)

jlamoreaux:
Third remediation PR. Addresses the **Critical scale cliff** the audit's performance pass found.

**Problem:** `listProjects`/`listProjectsByNamespace`/`listWorkspaces` each did an **unbounded** `Promise.all(keys.map(kv.get))` — one KV subrequest per entry, no concurrency limit. Past ~1000 projects/workspaces this exceeds the Workers **1000-subrequest-per-invocation cap and throws**. Because merge, dashboard, backup, and sync all route through `listProjects`, the **entire write path fails instance-wide** at that threshold.

**Fix:** bound the fan-out to a pool of 25 (`mapBounded`). All entries still returned; only concurrency is capped. Removes the hard crash.

**Not** the whole story: the deeper issue is needing a full-instance *scan* at all (every merge resolves identity via `listProjects`). That's the KV→D1 identity tranche (T5) — this PR just removes the cap-exceeded crash in the meantime.

Test: 60 projects list fully with peak concurrency ≤ 25. Full suite green (1046).

### DIFF
--- a/src/storage/state.ts
+++ b/src/storage/state.ts
@@ -6,6 +6,28 @@ import { type Result, err, ok } from "../utils/result";
 const PROJECT_PREFIX = "project:";
 const WORKSPACE_PREFIX = "workspace:";
 
+// KV get-per-key fan-out must stay well under the Workers ~1000-subrequest cap.
+// An unbounded `Promise.all(keys.map(get))` throws once an instance passes ~1000
+// projects/workspaces, taking the whole write path down; bound it to a pool.
+const KV_FANOUT_LIMIT = 25;
+
+async function mapBounded<T, R>(
+  items: readonly T[],
+  limit: number,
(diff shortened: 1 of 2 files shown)


jlamoreaux:
@coderabbitai review

jlamoreaux:
Updated: this PR now **also closes the pagination finding**. `listProjectsByNamespace` and `listWorkspaces` were doing a single `kv.list()` (one ~1000-key page) with no cursor loop, so large namespaces/projects silently dropped entries past page 1. Extracted the cursor loop into a `listAllKeyNames` helper and applied it to all three list functions; the bounded fan-out now runs over the complete key set. Tests cover multi-page completeness.

PR #138, opened 2026-07-20
fix(audit): tranche 3a — unique provenance per merge commit

jlamoreaux:
Part of the merge de-dup finding (High). A concurrent double-merge could write **two provenance rows for one merge commit**, corrupting the audit trail.

This is the **DB-level, path-independent** backstop (holds regardless of which of the three merge paths runs):
- **migration 033** — dedupe existing `(change_id, commit_sha)` rows, then a UNIQUE INDEX on `(change_id, commit_sha)`. Scope includes `commit_sha` so a legitimate re-merge after a revert (different sha) still records.
- `recordProvenance` → `INSERT OR IGNORE`: the losing concurrent writer is tolerated (logged), not an error.

**Remaining T3 piece** (separate PR): the status-transition CAS across the three merge paths to also de-dup the `change.merged` **event** (double webhooks / post-merge). Filed as follow-up.

Test: two records for one (change, commit) → one row; re-merge with a new commit still records. Full suite green (1046).

### DIFF
--- a/migrations/033_provenance_unique.sql
+++ b/migrations/033_provenance_unique.sql
@@ -1,0 +1,17 @@
+-- Prevent duplicate provenance rows for a single merge commit. A concurrent
+-- double-merge of the same change (the merge path is not wrapped in a CAS on
+-- every backend) could reach recordProvenance twice and write two rows for one
+-- commit, corrupting the audit trail.
+--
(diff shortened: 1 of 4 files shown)


jlamoreaux:
@coderabbitai review

PR #139, opened 2026-07-20
fix(audit): tranche 4a — backup health signal + cron audit

jlamoreaux:
Addresses the **silent backup rot** finding (High). Backups could degrade to useless while still reading as `complete: true` — `_manifest.json` is written unconditionally even if every export failed, and retention then evicts the last good run.

- **`healthy` flag** in the run summary/manifest: true only when every D1 table exported ok, KV is whole, and zero repos failed. Completeness (manifest present) ≠ success.
- **Error log** when a run is unhealthy — the actionable operator signal.
- **Cron audit row** (`backup.run`) with the healthy flag — the scheduled path previously left no audit trail at all (only the manual route did).

Follow-ups noted in the commit: retention should prefer healthy runs; an external alert transport; full-depth/all-refs capture.

Test: healthy=true on a clean run, healthy=false when a repo export fails. Full suite green (1045).

### DIFF
--- a/src/backup/run-backup.ts
+++ b/src/backup/run-backup.ts
@@ -1,3 +1,4 @@
+import { recordAudit } from "../storage/audit";
 import { RUN_MANIFEST_KEY, pruneRuns, putBlob } from "../storage/backup-store";
 import { BACKUP_TABLES, exportTable } from "../storage/d1-backup";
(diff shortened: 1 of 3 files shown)


jlamoreaux:
@coderabbitai review

PR #140, opened 2026-07-20
fix(audit): tranche 4b — backup captures full history (not depth-50 shallow)

jlamoreaux:
Fixes a real **correctness bug** in the backup feature: `snapshotRepo` clones via `cloneRepo`, which hardcodes `depth: 50`. So a repo with >50 commits was snapshotted with its **older ancestors missing** — the pack isn't reachability-closed and **can't restore the repo past commit 50**. The "full reachable history" backup silently wasn't.

- `cloneRepo` gains a `fullHistory` option: **merges keep the fast shallow clone**, **backup clones the complete history**. `MAX_BACKUP_BYTES` still bounds it (oversize repos skip + mark the run unhealthy, per #139).

**Follow-up** (noted in commit): all-refs capture (non-`main` branches/tags) needs manifest + reconstruct + restore changes — larger, tracked separately.

The git-clone depth flag isn't unit-testable without a git server; the reachability-closure it feeds is covered by the existing repo-snapshot walk test. Full suite green (1045).

## Summary by CodeRabbit

* **Bug Fixes**
  * Repository snapshots now preserve the complete reachable Git history instead of limiting it to a shallow history.
  * Restoring from a snapshot can now reliably return the repository to its exact latest commit.
  * Existing snapshot handling remains unchanged, including skipping empty or oversized repositories.

### DIFF
--- a/src/backup/repo-snapshot.ts
+++ b/src/backup/repo-snapshot.ts
@@ -138,7 +138,11 @@ export async function snapshotRepo(
   const token = await freshRepoToken(env.ARTIFACTS, project.remote, "read", logger);
   if (!token.success) return err(token.error);
 
-  const clone = await cloneRepo(project.remote, token.data, logger);
+  // Full history: a shallow clone would drop ancestors past depth 50, yielding a
+  // pack that can't restore the repo to its true tip.
+  const clone = await cloneRepo(project.remote, token.data, logger, undefined, {
+    fullHistory: true,
+  });
   if (!clone.success) return err(clone.error);
 
   const walk = await walkRepoObjects(clone.data.fs, clone.data.dir, maxBytes, logger);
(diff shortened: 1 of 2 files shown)


jlamoreaux:
Addressed in 76c3e22. This is a genuine but **inherent trade-off**, so I documented the rationale in code rather than attempting a bounded-clone rewrite:

- Full history is *required* for a restorable backup (a depth-50 shallow clone can't restore the true tip — the whole point of this PR).
- isomorphic-git's smart-HTTP fetch gives no way to know a repo's size **before** fetching, so there's no correctness-preserving *pre-clone* guard; the `MAX_BACKUP_BYTES` cap is necessarily post-clone.
- A bounded/streaming fetch that aborts mid-clone is the real fix — noted as a follow-up. In the meantime `MAX_BACKUP_BYTES` should be set well under the Worker memory budget so a normal repo never approaches it.

Marking as a known, documented trade-off.

PR #144, opened 2026-07-20
fix(ci): apply D1 migrations with --remote + reconciliation runbook (G1, Critical)

jlamoreaux:
## The bug (Critical)

CI applied D1 migrations **without `--remote`**, so `wrangler d1 migrations apply` targeted the runner's throwaway **local** SQLite instead of the Cloudflare-hosted database. The remote staging and production `d1_migrations` tables never recorded anything CI "applied." Fixed at all three sites:

- `.github/workflows/ci.yml:104` — staging deploy job
- `.github/workflows/ci.yml:168` — production deploy job (behind the `production` environment approval gate)
- `.github/workflows/d1-migrate.yml:56` — manual `workflow_dispatch`

(`pr-preview.yml` already had `--remote` — correct, it targets an ephemeral preview DB.)

## ⚠️ Do not merge until reconciliation is complete

`wrangler d1 migrations list DB --remote --env=production` shows production's tracking table is at **011**, with **012–032 (22 migrations) pending**. The first `--remote` apply would attempt all of them.

Re-apply-safety analysis of that backlog:
- Every `CREATE TABLE`/`CREATE INDEX` is `IF NOT EXISTS` → safe no-op if present.
- **Many carry `ALTER TABLE ADD COLUMN`** (013, 020, 024b, 025×7, 026–030, 032). SQLite/D1 has no `IF NOT EXISTS` for columns → re-adding an existing column errors `duplicate column name`.

Since app.usestratum.dev is live and needs this schema, it was almost certainly applied **out-of-band** and only the *tracker* is behind. A blind `--remote apply` would then sail through 012, **fail at 013**, and leave the tracker half-baselined. So the correct fix is to **baseline** the tracker (record 012–032 as applied without running them), *then* land this `--remote` change.

`docs/runbooks/d1-migration-reconciliation.md` (added here) documents:
- the diagnostic that discriminates *baseline* vs *staged-apply*,
- **Procedure A (baseline)** — verify schema + `INSERT OR IGNORE INTO d1_migrations`,
- **Procedure B (staged apply)** — staging first, verify, then production,
- the merge gate: complete reconciliation for **both** envs (list reports "No migrations to apply") **before merging this PR**.

Also flags the **duplicate `024`** (`024_backup_state.sql` + `024_change_workspace_head_sha.sql`) — left as-is (renaming an applied migration forces a re-run); track a rename for the next schema change.

## Scope

CI + runbook only — no `src`/test changes, so the suite is unaffected. The reconciliation writes (the `d1_migrations` inserts / applies) are operator-run against production per the runbook; this PR does not execute them.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,7 +101,7 @@ jobs:
           CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
         run: |
           echo "Applying D1 migrations for staging..."
-          npx wrangler d1 migrations apply stratum-staging --env=staging
+          npx wrangler d1 migrations apply stratum-staging --remote --env=staging
           echo "✓ Migrations applied successfully"
 
       - name: Deploy to Cloudflare Staging
@@ -165,7 +165,7 @@ jobs:
           CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
         run: |
           echo "Applying D1 migrations for production..."
-          npx wrangler d1 migrations apply stratum --env=production
+          npx wrangler d1 migrations apply stratum --remote --env=production
           echo "✓ Migrations applied successfully"
 
       - name: Deploy to Cloudflare Production
--- a/.github/workflows/d1-migrate.yml
+++ b/.github/workflows/d1-migrate.yml
@@ -53,7 +53,7 @@ jobs:
           fi
 
           echo "Applying migrations for environment=$ENV database=$DB_NAME"
-          npx wrangler d1 migrations apply "$DB_NAME" --env="$ENV"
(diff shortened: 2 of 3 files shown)


jlamoreaux:
**Diagnostics run (2026-07-20): it's the STAGED-APPLY path, not baseline.**

- **Production**: `SELECT … FROM sqlite_master` for webhooks/issues/audit_log/deletion_jobs/backup_state returned **zero rows** → those tables don't exist. Prod schema is genuinely at ~011; 012–032 were never applied (CI applied them locally-only — this bug). → **apply**, do NOT baseline (baselining would mark never-applied migrations as done and lock the schema out).
- **Staging**: 9 pending (`024_backup_state`, `025`–`032`); it applied `024_change_workspace_head_sha` but not `024_backup_state` — the duplicate-024 split its tracker state.

Procedure to follow from the runbook is **B (staged apply)**: export prod → apply staging → smoke-test → apply prod → then merge this PR. Follow-up: dedupe the two `024_*` files in a later migration-hygiene change.

PR #147, opened 2026-07-20
fix(audit): scope issue reads/updates by project_id + bound the list (T5)

jlamoreaux:
## What & why

`listIssues` / `getIssueByNumber` / `updateIssue` matched on the free-form `project` **name**, so a project could read (and update) a **same-named project's issues in another namespace** — the same cross-tenant leak class fixed for webhooks in #142.

Add an optional `projectId`; when provided, match:

```sql
WHERE (project_id = ? OR (project_id IS NULL AND project = ?)) …
```

Canonical id first, name only as a **legacy fallback** for rows whose `project_id` wasn't backfilled. Every route + UI caller (which already resolves the project for authz) now passes `project.id`.

## Deliberately scoped OUT (needs a gated migration)

This is the **read/update** isolation only. The hazardous parts of the KV→D1 identity migration are **not** here:
- switching issue-**number** generation to `MAX(number)+1 WHERE project_id=?`, and
- backfilling `project_id` onto legacy `NULL` rows.

Doing the numbering switch while legacy issues carry `NULL project_id` would **restart numbering at 1 and collide** with existing issue numbers — it needs a backfill-first, staged migration (its own `/ship` cycle with sign-off). This PR is intentionally non-breaking so it can land independently.

## Why it's safe / non-breaking

- Correctly-stamped rows return **identically** (project_id matches).
- Legacy `NULL`-project_id rows stay reachable via the name fallback.
- Number **generation is untouched**, so no renumber, no collision.
- New optional param → all existing storage-level callers/tests compile and pass unchanged.

## Tests

- Two same-named projects with distinct ids don't see each other's issues (`listIssues` + `getIssueByNumber`).
- A number that belongs to project A is **not** returned when scoped to project B.
- Legacy `NULL`-project_id rows remain reachable via the fallback.

Full suite **1053 passing**; typecheck + biome clean.

### DIFF
--- a/src/routes/issues.ts
+++ b/src/routes/issues.ts
@@ -28,6 +28,10 @@ const app = new Hono<{ Bindings: Env }>();
 const MAX_TITLE_LENGTH = 200;
 const MAX_BODY_LENGTH = 20_000;
 
+/** Default + hard cap for the paginated issues listing (bounds the response). */
+const DEFAULT_ISSUES_PAGE = 100;
+const MAX_ISSUES_PAGE = 500;
+
 interface RouteContext {
   env: Env;
   get(key: "userId" | "agentId" | "agentOwnerId"): string | undefined;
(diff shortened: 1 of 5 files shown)


jlamoreaux:
Updated: this PR now **also closes the unbounded-`listIssues` finding** — added an optional `limit` (default unlimited; internal callers unchanged) and the GET issues route caps the response at 100 (up to 500 via `?limit=`). Tests cover the bound.

PR #148, opened 2026-07-20
fix(audit): scope listChanges by project_id + bound the response (T5)

jlamoreaux:
## What & why

`listChanges` matched on the free-form `project` **name**, so a project could list a **same-named project's changes in another namespace** — the same cross-tenant read leak fixed for webhooks (#142) and issues (#147).

Add an optional `projectId`; when provided, match `(project_id = ? OR (project_id IS NULL AND project = ?))` — canonical id first, name only as a **legacy fallback**. All three callers (the changes list route + two UI pages, which already resolve the project) pass `project.id`.

## Safe / non-breaking

- Correctly-stamped rows return **identically**; legacy `NULL`-project_id rows stay reachable via the name fallback.
- `getChange` is keyed on the **globally-unique** change id (`chg_…`), so it was never cross-tenant — only the by-name *list* needed scoping.
- The numbering/backfill migration remains **out of scope** (gated), as with #147.

## Tests

- Same-named collision (two projects named `acme`, distinct ids) → each `listChanges` returns only its own changes.
- Legacy `NULL`-project_id rows reachable via the fallback.
- Name-only path (no `projectId`) unchanged.

Full suite **1053 passing**; typecheck + biome clean.

### DIFF
--- a/src/routes/changes.ts
+++ b/src/routes/changes.ts
@@ -126,6 +126,10 @@ function parseGitHubRepo(url: string): { owner: string; repo: string } | null {
 
 const MERGEABLE_STATUSES: Change["status"][] = ["approved", "accepted", "promoted"];
 
+/** Default + hard cap for the paginated changes listing (bounds the response). */
+const DEFAULT_CHANGES_PAGE = 100;
+const MAX_CHANGES_PAGE = 500;
+
 /**
  * Success response for endpoints that the change-detail UI posts to with plain HTML forms.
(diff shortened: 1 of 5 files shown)


jlamoreaux:
Updated: this PR now **also closes the unbounded-`listChanges` finding** — added an optional `limit` (default unlimited, so internal callers are unchanged) and the GET changes route caps the response at 100 (client may request up to 500 via `?limit=`). Tests cover the bound.

PR #204, opened 2026-08-18
Webhook evaluator: validate the verdict fail-closed and stop shipping webhook.secret in the payload

### LINKED ISSUE #195: Webhook evaluator verdict is unvalidated and leaks webhook.secret in the request payload
## Concern

`src/evaluation/webhook-evaluator.ts` trusts the external CI response without validation, and over-shares the policy.

- The response is cast, never checked: `const json = (await response.json()) as { score, passed, reason }`. A CI returning `{"passed":"no"}` (truthy string), `{"score":999}`, or omitting fields passes the gate — no clamping, no type check, no fail-closed. (Contrast the LLM evaluator, which does all three.)
- The request body is `JSON.stringify({diff, policy})` — it ships the **entire policy including `webhook.secret`** to the endpoint. The LLM path strips secrets via `sanitizePolicy`; the webhook path does not.

### Related architectural limit
The webhook evaluator is synchronous request/response only — there is no async callback API for a long external CI run to report back, so long runs must hold the HTTP connection open *inside change creation* on a Worker.

## Suggested direction
Validate/clamp the verdict and fail closed on malformed responses; strip secrets from the outgoing policy; consider an async status-callback path for external CI.

jlamoreaux:
## Summary

The webhook evaluator trusted the external CI endpoint's response verbatim (`await response.json() as {score, passed, reason}`), so `{"passed":"no"}` (a truthy string), `{"score":999}`, missing fields, or a non-JSON body could open the merge gate. It also serialized the full policy — including `webhook.secret` — into the outgoing request body.

Fix, mirroring the LLM evaluator's conventions:

- **Fail-closed validation:** the response must be a JSON object with a real `boolean` `passed` and a finite `number` `score` (clamped to [0,1]); `reason` defaults when not a string. Non-JSON bodies, non-object JSON, and missing/mistyped fields all return `{score: 0, passed: false}` with a warn log — the gate never passes on an unreadable verdict. Network/timeout errors still surface as `ExternalServiceError` as before.
- **Secret stripping:** the outgoing body is now `JSON.stringify({diff, policy: sanitizePolicy(policy)})`, reusing the same `sanitizePolicy` the LLM path uses — extracted from `llm-evaluator.ts` into a shared `src/evaluation/sanitize-policy.ts` (moved verbatim; LLM behavior unchanged) to avoid a webhook→llm import. The secret still signs the request via `X-Stratum-Signature`, computed over the sanitized body.

Notes for review:
- No threshold is applied to the webhook verdict (webhook config has none); `passed` is taken as reported once validated.
- A body-read timeout during `response.json()` now fails closed as "not valid JSON" instead of surfacing as an `ExternalServiceError` — consistent with fail-closed intent, but a slight behavior shift for that narrow race.
- The async status-callback path for long CI runs mentioned in the issue is out of scope here.

## Related issues

Closes #195

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## How was this verified?

12 new WebhookEvaluator tests in `tests/evaluation.test.ts`: valid verdict passthrough with clamping, truthy-string `passed` fails closed, non-finite/missing score, missing fields, non-JSON and non-object responses, and an assertion that the outgoing fetch body does not contain the secret. Coverage on `src/evaluation/webhook-evaluator.ts`: 100% statements/lines/functions, 97.43% branches (the one missed branch is v8's synthetic exceptional-finally entry, unreachable because the preceding catch is exhaustive); `sanitize-policy.ts` at 100%. Full suite: 1249 passed.

- [x] `npm run typecheck`
- [x] `npm test`
- [x] `npm run lint`

## Checklist

- [x] Docs updated where the change makes them inaccurate
- [x] The web UI change (if any) stays server-rendered with no client-side JavaScript

---
_Generated by [Claude Code](https://claude.ai/code/session_015dp67PBbFT666GnMLuxvm1)_

## Summary by CodeRabbit

* **Security**
  * Webhook requests no longer transmit evaluator secrets, while signing continues to work securely.

* **Bug Fixes**
  * Improved handling of malformed, invalid, or unexpected webhook responses.
  * Invalid scores are safely constrained to the 0–1 range.
  * Missing or invalid result details now receive safe default values.
  * Webhook evaluation failures now fail closed with a score of zero.

### DIFF
--- a/src/evaluation/llm-evaluator.ts
+++ b/src/evaluation/llm-evaluator.ts
@@ -4,7 +4,8 @@ import { ExternalServiceError } from "../utils/errors";
 import type { Logger } from "../utils/logger";
 import type { Result } from "../utils/result";
 import { err, ok } from "../utils/result";
-import type { EvalPolicy, EvalResult, Evaluator, EvaluatorConfig } from "./types";
+import { sanitizePolicy } from "./sanitize-policy";
+import type { EvalPolicy, EvalResult, Evaluator } from "./types";
 
 const DEFAULT_MODEL = "@cf/meta/llama-3.1-8b-instruct";
 const DEFAULT_THRESHOLD = 0.7;
@@ -27,19 +28,6 @@ const SYSTEM_PROMPT = [
   "Respond with ONLY a JSON object and no other text:",
   '{"score": <0.0-1.0>, "passed": <bool>, "reason": "<one sentence>", "issues": ["<finding>"]}',
 ].join(" ");
-
-function sanitizePolicy(policy: EvalPolicy): EvalPolicy {
-  return {
-    ...policy,
-    evaluators: policy.evaluators.map((cfg: EvaluatorConfig) => {
-      if (cfg.type === "webhook") {
-        const { secret: _secret, ...rest } = cfg;
-        return rest;
-      }
-      return cfg;
-    }),
-  };
-}
 
 export class LLMEvaluator implements Evaluator {
   constructor(private ai: AiBinding) {}
--- a/src/evaluation/sanitize-policy.ts
+++ b/src/evaluation/sanitize-policy.ts
@@ -1,0 +1,19 @@
+import type { EvalPolicy, EvaluatorConfig } from "./types";
+
+/**
+ * Strip credentials from a policy before it leaves the Worker (serialized into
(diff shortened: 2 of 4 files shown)


jlamoreaux:
@coderabbitai review

---

Issue #235, opened 2026-08-19
Backfill webhooks.project_id and remove the name-based tenant fallback


jlamoreaux:
## Background

PR #227 scoped webhook management by `project_id` and stopped leaking signing secrets. It intentionally kept the legacy name fallback in `webhookBelongsToProject` and `listWebhooks`:

```
project_id = ? OR (project_id IS NULL AND project = ?)
```

Webhooks created **before migration 025** (which added `webhooks.project_id`) have a null `project_id`. For those legacy rows, the name fallback still permits a same-named project in another namespace to match — across the management, listing, **and delivery** paths (the delivery consumer `src/queue/webhook-delivery.ts` shares `listWebhooks`). Every webhook created since 025 carries `project_id`, so this affects legacy rows only.

Removing the fallback in isolation is unsafe: it would orphan legitimate legacy webhooks on single-tenant instances (owner can no longer manage them) and wouldn't fully close the delivery path.

## The fix (coordinated)

1. **Backfill** `webhooks.project_id` for all null rows. Project identity lives in KV, so this is an **app-level** pass (resolve each row's `project` name → KV project id → `UPDATE webhooks SET project_id = ?`), not a pure SQL migration. Model it on the existing admin backfill endpoints.
2. **Verify** no null `project_id` rows remain (a check/report before step 3).
3. **Remove the name fallback** consistently:
   - `webhookBelongsToProject` (`src/storage/webhooks.ts`) — compare `project_id` only.
   - `listWebhooks` — drop the `OR (project_id IS NULL AND project = ?)` clause (affects both management and delivery callers).
   - Update `tests/webhooks.test.ts` / `tests/webhooks-routes.test.ts` to assert legacy null-`projectId` rows are rejected rather than name-matched.

**Acceptance:** verification confirms zero null `project_id` rows in `webhooks`, and no name-based fallback remains in the helper, `listWebhooks`, or the delivery query.

Tracks the residual raised by CodeRabbit on #227.

PR #268, opened 2026-08-24
Admin apply-endpoint to backfill webhooks.project_id from KV (steps 1–2 of #235)

jlamoreaux:
## Summary

Implements the **backfill** and **verification** halves of #235's coordinated fix. `webhooks` rows written before migration 025 carry a NULL `project_id`; project identity lives in KV, so this is an app-level pass — exactly what `scripts/backfill-project-id.ts` documented as the intended shape but never implemented for lack of an admin write endpoint (the existing `/api/admin/backfill-project-id/plan` is read-only).

- **New:** `POST /api/admin/backfill-project-id/webhooks/apply` — admin-only, mirroring the existing plan endpoint's `requireAdmin` pattern. For every NULL-`project_id` row it resolves the row's `project` name against the KV project directory and stamps `project_id`, but **only when the name resolves to exactly one project across all namespaces**. Ambiguous or unresolved names are left NULL and reported — never guessed, since this is a tenant-isolation boundary and a bare name can collide across namespaces.
- The response reports `{ updated, skipped: [{ webhookId, project, reason }], remainingNullRows }` — `remainingNullRows` is the step-2 verification the issue calls for; it should read `0` after a clean run.
- Writes an audit entry (`webhook.project_id_backfilled`) per run. Updates re-assert `project_id IS NULL` at write time as defense against concurrent runs.

**Step 3 (removing the name-based fallback in `webhookBelongsToProject` / `listWebhooks`) is deliberately NOT in this PR.** It's only safe after this backfill has run against production and the report confirms zero remaining NULL rows; removing it earlier would orphan any legacy webhook whose name is ambiguous/unresolved. That removal should follow as its own PR once an operator has run the apply endpoint.

## Related issues

Refs #235 (steps 1–2; fallback removal deferred until production backfill is verified)

## Type of change

- [x] New feature

## How was this verified?

- [x] `npm run typecheck`
- [x] `npm test` (full suite: 1729 passed, 0 failures)
- [x] `npm run lint`

- New unit tests (`tests/backfill-plan.test.ts`): unique-name row backfilled; ambiguous name (two namespaces, same project name) stays NULL and is reported; unresolved name stays NULL and is reported; already-backfilled rows untouched.
- New route tests (`tests/backfill-route.test.ts`): 403 without admin credentials (backfill/audit never invoked); 200 + audit entry for an admin.

## Checklist

- [x] Docs updated where the change makes them inaccurate (script stub already describes this shape)
- [x] The web UI change (if any) stays server-rendered with no client-side JavaScript (no UI change)

https://claude.ai/code/session_01Gk1wNDsuwpt6q36oZaZhof

---
_Generated by [Claude Code](https://claude.ai/code/session_01Gk1wNDsuwpt6q36oZaZhof)_

## Summary by CodeRabbit

* **New Features**
  * Added a protected endpoint to apply webhook project associations.
  * Webhooks are updated only when their project name uniquely matches a project.
  * Results include updated, skipped, unresolved, and remaining-unmatched records.
  * Added audit logging with an indicator showing whether auditing succeeded.
  * Added documentation for the administrative operation.

* **Bug Fixes**
  * Prevented overwriting existing associations and handled concurrent updates safely.
  * Ambiguous or unresolved matches remain unchanged.
  * Audit failures no longer prevent successful webhook updates.

### DIFF
--- a/src/routes/backfill.ts
+++ b/src/routes/backfill.ts
@@ -8,6 +8,11 @@ import { forbidden, internalError, ok } from "../utils/response";
 
 const app = new Hono<{ Bindings: Env }>();
 
+/**
+ * Gate an admin-only backfill route, returning a request-scoped logger on
+ * success. Accepts either the X-Admin-API-Key header or an authenticated
+ * admin user, matching the existing /plan endpoint's contract.
+ */
 async function requireAdmin(c: {
   env: Env;
   req: { header: (n: string) => string | undefined; path: string; method: string };
@@ -64,7 +69,7 @@ app.post("/webhooks/apply", async (c) => {
   const report = await backfillWebhookProjectIds(c.env, logger);
   if (!report.success) return internalError(report.error.message);
 
-  await recordAudit(c.env.DB, logger, {
+  const auditResult = await recordAudit(c.env.DB, logger, {
     action: "webhook.project_id_backfilled",
     actorType: c.get("userId") ? "user" : "system",
     ...(c.get("userId") !== undefined ? { actorId: c.get("userId") } : {}),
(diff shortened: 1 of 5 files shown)


jlamoreaux:
@coderabbit review

PR #287, opened 2026-08-26
Backfill route: audit partial runs, and attribute audits to the credential that authorized them

jlamoreaux:
## Summary

Follow-up to #268, which merged before two review findings on it could land. Both are auditability defects in `POST /api/admin/backfill-project-id/webhooks/apply`, verified against `main` at `81704f8`:

**1. A partially-applied backfill left no audit trail.** `backfillWebhookProjectIds` issues one `UPDATE` per row, and D1 auto-commits each independently — there is no enclosing transaction. A throw on row N therefore leaves rows 1..N-1 already stamped, but the route returned `internalError(...)` *before* ever reaching `recordAudit`. A privileged mutation landed with zero provenance.

The backfill is deliberately idempotent and resumable, so partial application is expected and fine; doing it invisibly is not. Rather than force the whole run into one D1 `batch()` (which would put an unbounded number of statements in a single batch and defeat resumability), the function now carries its partial `updated`/`skipped` counts out on the returned `AppError`'s `context` — mirroring how `applySourceUpdate` attaches `context: { missingMergeBase }` — and the route records a `failed: true` audit entry from them before returning the 500. That write is best-effort, exactly like the success path: if it also fails, it is logged and the original error is returned untouched.

**2. API-key-authorized runs were attributed to the wrong actor.** `isAdminRequest` checks the `X-Admin-API-Key` branch first and returns without ever consulting `userId`, but the route derived `actorType` from `c.get("userId") ? "user" : "system"`. A service-to-service run that happened to carry any user session — including a non-admin one — recorded that user as the actor. Misattribution in an audit trail is worse than an absent `actorId`.

Added `resolveAdminAuth`, which returns which credential granted access. `isAdminRequest` becomes a thin boolean wrapper over it, so its **signature is unchanged** and its other five call sites (`audit.ts`, `backup.ts`, `deletion-jobs.ts`, `metrics.ts`, `restore.ts`) are untouched. Branch order, checks, and the fail-closed-when-neither-secret-is-configured behavior are all identical — only the reported attribution changed.

The `audited: false` success-path contract agreed on in #268 is unchanged: a failed audit write after a successful backfill is still a 200 with `audited: false`, not a 500, since `recordAudit` is best-effort by contract and a 500 would discard the `remainingNullRows` count the endpoint exists to report.

## Related issues

Follow-up to #268; part of #235 (steps 1–2).

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How was this verified?

- `npm run typecheck` — clean.
- `npm run lint` — clean, 302 files.
- `npx vitest run` (full suite) — 141 files passed, 3 skipped; **2130 tests passed, 27 skipped, 0 failures**.
- New tests in `tests/backfill-route.test.ts`:
  - failure path records the partial `updated`/`skipped` with `failed: true` and still returns 500;
  - a failure-path audit write that itself fails still returns the original 500 and message untouched;
  - mixed credentials (valid `X-Admin-API-Key` + a non-admin `userId` attached by middleware) record `actorType: "system"` with no `actorId`.
- Re-ran the suites for every other `isAdminRequest` call site unmodified (`audit`, `metrics`, `backup`, `restore`, `deletion-jobs`, `deletion-routes`) — 46/46 pass.

## Checklist

- [x] Tests added/updated for the change
- [x] Full test suite passes locally
- [x] Lint and typecheck pass
- [x] No changes to public API signatures (`isAdminRequest` is unchanged; `resolveAdminAuth` is additive)

---

https://claude.ai/code/session_01Gk1wNDsuwpt6q36oZaZhof

---
_Generated by [Claude Code](https://claude.ai/code/session_01Gk1wNDsuwpt6q36oZaZhof)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved admin authorization handling for backfill operations.
  * Backfill failures now preserve the original error details and report partial progress.
  * Failed operations now receive audit records, including cases where only some items were processed.
  * Audit records more accurately identify whether an operation was authorized by an API key or user credentials.
  * Preserved the original operation error even when failure auditing encounters an additional error.

### DIFF
--- a/src/routes/backfill.ts
+++ b/src/routes/backfill.ts
@@ -2,7 +2,7 @@ import { Hono } from "hono";
 import { recordAudit } from "../storage/audit";
 import { backfillWebhookProjectIds, computeBackfillPlan } from "../storage/backfill-plan";
 import type { Env } from "../types";
-import { isAdminRequest } from "../utils/admin";
+import { resolveAdminAuth } from "../utils/admin";
 import { createLogger } from "../utils/logger";
 import { forbidden, internalError, ok } from "../utils/response";
 
@@ -12,19 +12,26 @@ const app = new Hono<{ Bindings: Env }>();
  * Accepts either the X-Admin-API-Key header or an authenticated admin user,
  * deliberately matching /plan's contract: an operator running the dry-run and
  * the apply back-to-back should not need two different credentials.
+ *
+ * Carries `via` out alongside `ok` so callers that write an audit entry can
+ * attribute the action to whichever credential actually authorized it,
+ * rather than to a `userId` that merely happened to be attached to the
+ * request (see resolveAdminAuth).
  */
 async function requireAdmin(c: {
   env: Env;
   req: { header: (n: string) => string | undefined; path: string; method: string };
   get: (k: "userId") => string | undefined;
-}): Promise<{ ok: true; logger: ReturnType<typeof createLogger> } | { ok: false }> {
+}): Promise<
+  { ok: true; logger: ReturnType<typeof createLogger>; via: "api-key" | "user" } | { ok: false }
+> {
   const logger = createLogger({
     requestId: crypto.randomUUID(),
     userId: c.get("userId"),
     path: c.req.path,
     method: c.req.method,
   });
-  const isAdmin = await isAdminRequest(
+  const auth = await resolveAdminAuth(
     c.env,
     {
       ...(c.req.header("X-Admin-API-Key") !== undefined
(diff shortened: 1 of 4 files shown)


jlamoreaux:
@coderabbitai review

The one finding from the previous review (type docs on `src/utils/admin.ts`) was fixed in `ae7724b` and you marked it addressed, and the thread is resolved — but the incremental review of that commit was rate-limited, so the review state still reads changes-requested. CI is green on all 11 checks and the branch merges cleanly into `main`.

---

Issue #289, opened 2026-08-27
GitHub client retries without settling the response body, retaining Workers connection resources


jlamoreaux:
## Summary

`GitHubClient.request()` retries on 403 rate-limit and on 5xx by awaiting a delay and recursing — without reading or cancelling the `Response` it already holds. In Cloudflare Workers an unconsumed response body keeps its connection resources held, so repeated retries can stall or exhaust the pool.

## Where

`src/github/client.ts`, on `claude/issue-243-244-promotion-evaluated-sha` at `5504f85` (PR #275).

The 403 rate-limit branch, `client.ts:390-397`:

```ts
await new Promise((resolve) => setTimeout(resolve, Math.max(waitTime, 1000)));
return this.request(endpoint, options, { ...requestOpts, retryCount: retryCount + 1 });
```

`response` is in scope and never settled. The 5xx retry branch has the same shape.

## Why it matters

The retry path is exactly the path that runs repeatedly, and it is reached under the two conditions where GitHub is already degraded — rate limiting and server errors. That is when connection pressure is least affordable.

Bounded by `maxRetries`, so this is resource pressure rather than a leak without limit. But every retried request holds a connection it no longer needs for the whole backoff delay, which on a rate-limit wait can be seconds.

## Suggested direction

Cancel or drain the body before the delay in both branches — `await response.body?.cancel()` (or read and discard) ahead of the `setTimeout`. Cancelling is preferable to draining: the bytes are not wanted and cancelling releases immediately.

Worth a test that drives repeated 5xx responses with bounded retries and asserts the body is settled on each attempt, since this is invisible to any assertion about the final result.

## Provenance

Found by CodeRabbit reviewing #275 (run `abf65850`), as an outside-diff-range comment. Verified against the code at `5504f85` before filing.

Issue #288, opened 2026-08-27
Duplicate-head 422 counts against the GitHub circuit breaker, so PR reuse can be blocked by the failures it exists to recover from


jlamoreaux:
## Summary

`GitHubClient.createOrReusePR` treats a duplicate-head `422` as an expected, recoverable outcome — it falls through to a lookup that finds the existing PR and reuses it. But `request()` has already recorded that `422` as a **circuit-breaker failure**, and the lookup goes through the same breaker.

So the mechanism that recovers from duplicate-head is counted as a failure by the mechanism that can block it.

## Where

`src/github/client.ts`, on `claude/issue-243-244-promotion-evaluated-sha` at `5504f85` (PR #275):

- `client.ts:430` — `recordCircuitResult(endpoint, false)` on the generic non-ok path. A `422` lands here.
- `client.ts:569-580` — the duplicate-head branch, which then calls `this.request(...)` for `/pulls?head=…` through the same client.
- The breaker is keyed per repository (`/repos/<owner>/<repo>`), threshold 5 consecutive failures, `CIRCUIT_BREAKER_TIMEOUT_MS` cooldown.

## Why it matters

Re-promoting the same change is the normal way to hit duplicate-head, so these `422`s are routine rather than exceptional. Five of them against one repository — sequentially or concurrently — open the shared breaker. The reuse lookup then returns `503 Service temporarily unavailable (circuit open)`, `candidate` is `undefined`, and `createOrReusePR` falls through to reporting the original `422` as a hard failure.

The result is that a benign "PR already exists" becomes a failed promotion, for the duration of the cooldown, precisely when the user is retrying.

It is bounded and self-healing — the breaker closes on its own — but the failure is user-visible and the trigger is a routine action.

## Suggested direction

Either:

- exclude the duplicate-head `422` from breaker accounting at the point it is recognised as expected, or
- give the breaker a policy limited to the failures it is actually meant to shed — transport errors, 5xx, and rate limiting — so that 4xx responses carrying a meaningful API answer do not count.

The second is the more durable shape: a `4xx` generally means GitHub answered successfully and the caller should interpret it, which is not the condition a breaker exists for.

Worth a regression test that drives repeated duplicate-head creates and asserts reuse still succeeds.

## Scope note

The circuit breaker itself predates PR #275 — that PR routes one additional caller through it. What is new is that this caller *routinely* produces a response the breaker counts as a failure, which is what turns a latent design gap into a reachable one.

## Provenance

[...]

PR #292, opened 2026-08-27
Search every duplicate-head candidate in createOrReusePR, not just the first

### LINKED ISSUE #290: createOrReusePR only considers the first duplicate-head candidate, so a reusable PR can be missed
## Summary

After a duplicate-head `422`, `createOrReusePR` looks up open PRs for the head and then tests **only the first result** against the caller's predicate. If that one is rejected but a later one would be accepted, the method discards the lookup and reports the original `422` instead of reusing the PR that was there all along.

## Where

`src/github/client.ts:581`, on `claude/issue-243-244-promotion-evaluated-sha` at `5504f85` (PR #275):

```ts
const candidate = lookup.success && Array.isArray(lookup.data) ? lookup.data[0] : undefined;
if (candidate !== undefined && isReusablePr(candidate)) {
  return { success: true, pr: candidate, reused: true, status: lookup.status };
}
```

## Why it matters

The lookup is `GET /pulls?head=<owner>:<head>&state=open`, which is head-only and can legitimately return more than one open PR. And the predicate is not a formality — the production caller passes `isUsableGithubPr`, which rejects PRs that do not belong to the repository the promotion is actually pushing to (a project migrated between GitHub repos, or `sourceUrl` superseding a legacy `githubUrl`).

So the case where `[0]` is rejected and a later entry qualifies is exactly the case the predicate exists to handle. Taking only `[0]` means the reuse path silently fails in the situation it was written for.

Failure mode is a confusing error rather than corruption: the user sees "pull request already exists" while an open, reusable PR sits there.

## Suggested direction

```ts
const candidate = lookup.success && Array.isArray(lookup.data)
  ? lookup.data.find(isReusablePr)
  : undefined;
```

which folds the predicate into the search rather than applying it after an arbitrary pick. Worth a test with two open PRs where the first is rejected and the second accepted.

## Provenance

jlamoreaux:
## Summary

After a duplicate-head `422`, `createOrReusePR` looked up the open PRs for the head and tested **only `lookup.data[0]`** against the caller's predicate. If the first entry was rejected but a later one would have been accepted, the lookup was discarded and the original `422` reported instead.

That matters because the lookup is `GET /pulls?head=:&state=open` — head-only, so more than one open PR is legitimate — and the production caller passes `isUsableGithubPr`, which rejects PRs belonging to a repository the promotion is not pushing to (a project migrated between GitHub repos, or `sourceUrl` superseding a legacy `githubUrl`). The case where `[0]` is rejected and a later entry qualifies is exactly the case the predicate exists to handle, so the reuse path silently failed in its own motivating scenario: the user saw "pull request already exists" while an open, reusable PR sat there.

The fix folds the predicate into the search rather than applying it after an arbitrary pick:

```ts
const candidates: unknown[] = lookup.success && Array.isArray(lookup.data) ? lookup.data : [];
const candidate = candidates.find(isReusablePr);
if (candidate !== undefined) { ... }
```

Typing the list as `unknown[]` also drops the implicit `any` that `Array.isArray` narrowing produced on the old `lookup.data[0]`, which `AGENTS.md` forbids. The `createOrReusePR` JSDoc is updated where it described the predicate as gating one hit.

**Fall-through is unchanged.** When the predicate accepts nothing — lookup failed, found nothing open, or rejected every entry — the original create error is still what gets reported. A regression test now guards that across a multi-entry response, not just a single-entry one.

### Explicitly not in scope

- **Pagination of `GET /pulls`.** The lookup is head-scoped and open-only; GitHub's default page of 30 is far beyond any plausible count for one head ref.
- **Circuit-breaker accounting on the duplicate-head `422` (#288)** and **response-body settling on retry (#289)** — filed separately, shipping separately.

## Related issues

Closes #290

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## How was this verified?

The two new cases were confirmed to be genuine regression tests, not just passing assertions: with the test file in place and `src/github/client.ts` reverted to its pre-fix state, `reuses a later lookup entry when the caller's predicate rejects the first` **fails**; it passes once the fix is restored.

- [x] `npm test` — 2195 passed, 27 skipped, 0 failed (145 files)

## Checklist

- [x] Docs updated where the change makes them inaccurate — swept `README.md`, `AGENTS.md`, `CONTRIBUTING.md` and `docs/` for `createOrReusePR`; it is internal and documented only in its own JSDoc, which this PR updates. No other docs affected.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pull request reuse when multiple matching open pull requests exist.
  * The system now selects the first eligible matching pull request instead of stopping at an ineligible result.
  * Preserved the original creation error when no matching pull request can be reused.

### DIFF
--- a/src/github/client.ts
+++ b/src/github/client.ts
@@ -519,15 +519,17 @@ export class GitHubClient {
    * against whatever they trust enough to store, since a JSON parse alone
    * doesn't guarantee GitHub's documented fields are actually present.
    *
-   * `isReusablePr` gates the lookup hit before it's treated as a successful
+   * `isReusablePr` selects the lookup hit that gets treated as a successful
    * reuse: this client has no opinion on what a caller trusts enough to
    * accept as "the PR" (the promotion route, for one, only accepts an open
    * PR whose `html_url` is exactly this repo's canonical page for the
-   * matched number — see `isUsableGithubPr`). A lookup hit the predicate
-   * rejects behaves exactly like a lookup that failed or found nothing open:
-   * the ORIGINAL create error is what gets reported, not a new failure about
-   * the lookup response, since that's the one call the caller actually asked
-   * for.
+   * matched number — see `isUsableGithubPr`). It is applied across every PR
+   * the lookup returns, not just the first, since the lookup is head-scoped
+   * and a head can carry more than one open PR. A lookup whose entries the
+   * predicate all reject behaves exactly like a lookup that failed or found
+   * nothing open: the ORIGINAL create error is what gets reported, not a new
+   * failure about the lookup response, since that's the one call the caller
+   * actually asked for.
    *
    * Required, deliberately not defaulted. The code this replaced always ran
    * an owner/repo check on the lookup hit, and a parameter defaulting to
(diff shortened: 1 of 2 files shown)


jlamoreaux:
@coderabbit review

PR #294, opened 2026-08-27
Keep request-scoped 4xx answers out of circuit-breaker accounting

### LINKED ISSUE #288: Duplicate-head 422 counts against the GitHub circuit breaker, so PR reuse can be blocked by the failures it exists to recover from

jlamoreaux:
## Summary

`createOrReusePR` treats a duplicate-head `422` as expected and recoverable — it falls through to a lookup that finds the existing PR and reuses it. But `request()` had already recorded that `422` as a circuit-breaker failure, and the reuse lookup goes through the same per-repository breaker. The mechanism that recovers from duplicate-head was counted as a failure by the mechanism that can block it.

This adds `isCircuitFailureStatus()`. The breaker is keyed **per repository**, so what it measures is whether that repository is reachable and usable. The policy follows from that:

| | Statuses | Why |
|---|---|---|
| **Counts** | transport errors, 5xx, rate limiting, `401`, `403`, `404`, `410` | The repository is unreachable, gone, or the credential cannot use it |
| **Neutral** | `400`, `409`, `422` | GitHub was reached, answered, and is describing *this request*. The caller interprets it. |

## Two things reviewers should weigh

**1. This is narrower than the issue's suggested policy, on purpose.** The issue proposes that 4xx responses generally should not count, and calls that the more durable shape. It is — but the literal reading collides with the breaker's own documented purpose. `isCircuitOpen`'s comment says it exists so that *"a repo that is gone, renamed, or whose token lost access"* is not rediscovered once per endpoint. Those are `404` and `401`/`403`. Excluding every 4xx would delete that behaviour, so `401`/`403`/`404` stay counted and a test now guards it. If maintainers prefer the literal reading, the change is one predicate — but `isCircuitOpen`'s doc comment would need updating to match.

**2. The issue's premise needs one refinement.** It says five duplicate-head `422`s "sequentially or concurrently" open the breaker. Sequentially, they do not: each round's reuse lookup returns `200`, and any success zeroes the consecutive-failure tally, so the count never climbs past one between rounds. I verified this — an early version of the regression test passed against the *unfixed* client for exactly that reason, which is what surfaced it.

The genuinely reachable shape is a duplicate-head `422` landing **on top of an existing failure run** against the same repository. The breaker is shared per repository, so any caller's failures count toward it — a burst of 404s from evaluation reporting leaves the breaker at 4, and the next promotion's `422` is failure #5. The lookup it recovers through is then short-circuited with `503`, and the promotion fails. That is what the headline test drives, and it is the case that fails without this fix.

The user-visible outcome and the fix are unchanged by this refinement; the trigger is somewhat narrower than the issue describes.

### Neutral, not success

A non-counting failure records neither outcome. Recording success would zero a genuine consecutive-failure run because one `422` landed in the middle of it, which is the opposite of what a consecutive counter is for. While half-open, skipping correctly leaves the breaker half-open — a probe that drew a request-scoped answer learned nothing about the repository.

### Explicitly not in scope

Threshold, cooldown, per-repository keying, and the half-open probe are unchanged. The duplicate-head candidate search (#290) and retry body settling (#289) are separate PRs.

## Related issues

Closes #288

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## How was this verified?

Five cases in `tests/github-client-circuit-breaker.test.ts`:

1. **Reuses on duplicate-head with the breaker one failure from opening** — the regression. Fails against the pre-fix client (`503`, circuit open); passes with the fix.
2. **A run of duplicate-head promotions never opens the breaker** — asserts two fetches per round throughout, so nothing was short-circuited.
3. **Repeated 404s still open the breaker** — guards the documented behaviour this PR deliberately preserves.
4. **A `422` is neutral, not a reset** — four genuine failures, then a `422`, then one more failure trips the threshold. Had the `422` cleared the tally, it would not.
5. **`422`s alone never open the breaker** — the caller always sees GitHub's `422`, never the breaker's `503`.

Cases 3 and 4 pass in both directions by design: they guard behaviour that must not change.

- [x] `npm test` — 2198 passed, 27 skipped, 0 failed (146 files)

## Checklist

- [x] Docs updated where the change makes them inaccurate — grepped `README.md`, `AGENTS.md`, `CONTRIBUTING.md` and `docs/` for circuit-breaker behaviour; the only descriptions are the JSDoc in `client.ts`, which this PR extends so the documented purpose and the code agree.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved circuit-breaker handling for expected GitHub responses.
  - Duplicate or validation-related responses no longer incorrectly open or alter repository circuit breakers.
  - Repeated not-found and other server failures continue to be tracked appropriately.
  - Prevented valid operations from returning circuit-open errors after non-failure responses.

- **Tests**
  - Added coverage for pull request operations near circuit-breaker thresholds.
  - Verified expected responses preserve breaker state while genuine failures continue to trigger protection.

### DIFF
--- a/src/github/client.ts
+++ b/src/github/client.ts
@@ -157,6 +157,37 @@ export function resetCircuitBreakersForTests(): void {
 }
 
 /**
+ * Statuses that GitHub uses to answer a *request* rather than to report on the
+ * *repository*.
+ *
+ * The breaker is keyed per repository, so what it measures is whether that
+ * repository is reachable and usable. A `422` saying "a pull request already
+ * exists for this head" is GitHub reached, GitHub answering, and the caller
+ * expected to interpret the answer — it says nothing about repository health,
+ * and it must not move a repository-health counter. That is not academic:
+ * duplicate-head `422`s are the ordinary result of re-promoting a change, and
+ * counting them let five routine re-promotions open the breaker that the
+ * duplicate-head RECOVERY path then has to pass through (see
+ * `createOrReusePR`), turning "PR already exists" into a failed promotion.
+ *
+ * Deliberately narrower than "no 4xx counts": `401`, `403`, and `404` stay
+ * counted, because a repository that is gone, renamed, or beyond the token's
+ * access is precisely the condition {@link isCircuitOpen} exists to stop
+ * rediscovering once per endpoint.
+ */
+const REQUEST_SCOPED_STATUSES = new Set([400, 409, 422]);
+
+/**
+ * Whether a failed response should feed its repository's breaker.
+ *
+ * Transport errors and rate limiting are recorded by their own call sites and
+ * never reach here; this decides only for responses carrying a status.
+ */
+function isCircuitFailureStatus(status: number): boolean {
+  return !REQUEST_SCOPED_STATUSES.has(status);
+}
+
+/**
  * Whether calls to `endpoint`'s repository are currently being short-circuited.
  *
  * Keyed by the first three path segments (`/repos/<owner>/<repo>`), so the
@@ -195,6 +226,14 @@ function isCircuitOpen(endpoint: string): boolean {
  * of failures, so an intervening success means whatever was wrong is no longer
  * reproducing and the tally should not carry forward into an unrelated later
  * incident.
+ *
+ * Not every failed call gets here. A response the caller is expected to
+ * interpret (see {@link isCircuitFailureStatus}) is NEITHER recorded as a
+ * failure nor as a success — it is skipped entirely. Recording it as a success
+ * would zero a genuine consecutive-failure run because one `422` landed in the
+ * middle of it, which is the opposite of what a consecutive counter is for; and
+ * while half-open, skipping correctly leaves the breaker half-open, since a
+ * probe that drew a request-scoped answer learned nothing about the repository.
  */
 function recordCircuitResult(endpoint: string, success: boolean): void {
   const key = endpoint.split("/").slice(0, 3).join("/");
@@ -427,7 +466,9 @@ export class GitHubClient {
           status: response.status,
           error: errorText,
         });
-        recordCircuitResult(endpoint, false);
+        if (isCircuitFailureStatus(response.status)) {
+          recordCircuitResult(endpoint, false);
+        }
         return {
           success: false,
           error: `GitHub API error: ${response.status} - ${errorText}`,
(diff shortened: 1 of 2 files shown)


jlamoreaux:
@coderabbit review

jlamoreaux:
@coderabbit review

PR #295, opened 2026-08-27
Record the published revision on promotion's stored-PR reuse path

### LINKED ISSUE #291: Promotion's stored-PR reuse path returns without persisting githubHeadSha, leaving the recorded published revision stale
## Summary

The promotion route computes and validates `publishedSha` against `change.evaluatedSha`, then force-pushes. But when the change already has usable stored PR metadata, the route returns early — before the `updateChangeStatus` call that writes `githubHeadSha`. The push succeeds and the published revision is known, and it is not recorded.

## Where

`src/routes/changes.ts`, on `claude/issue-243-244-promotion-evaluated-sha` at `5504f85` (PR #275). Three points, in execution order:

| Line | What happens |
|---|---|
| 1676-1688 | `publishedSha` is computed from the local tip and validated against `change.evaluatedSha` |
| ~1759-1774 | stored-PR reuse branch: `if (isUsableGithubPr(storedPr, repo) && storedPrMatchesTarget)` → `return okOrFormRedirect(...)` |
| 1898 | `...(publishedSha !== undefined ? { githubHeadSha: publishedSha } : {})` |

The early return at 1774 never reaches 1898.

## Why it matters

This is the *re-promotion* path — the one taken when a change is promoted again against an existing PR. So it is not an edge case; it is the second and every subsequent promotion of any change.

`githubHeadSha` is the record of which revision was actually published. After a re-promotion it can be missing (legacy rows) or stale (pointing at the previous head) while the branch on GitHub has moved. Anything reading that field to answer "what is currently published?" gets a wrong answer, and nothing surfaces the discrepancy.

Data-fidelity rather than correctness — the promotion itself succeeds and the right commits reach GitHub. What is wrong is the metadata describing it.

## Suggested direction

Persist the confirmed `publishedSha` on the reuse path before returning, including when the stored value is missing or stale. Care is needed not to disturb the surrounding validation: the early return exists because reusing a PR that matches the target is the correct answer, and that should stay — only the write is missing.

Worth a regression test covering re-promotion with valid stored PR metadata after a force-push, asserting `githubHeadSha` matches the new head rather than the old one.

## Provenance

Found by CodeRabbit reviewing #275 (run `abf65850`), as an outside-diff-range comment. I had not traced this one when it was first raised; verified against the code at `5504f85` before filing.

jlamoreaux:
## Summary

The promotion route computes `publishedSha`, validates it against `change.evaluatedSha`, and force-pushes — then, when the change already has usable stored PR metadata, returns early, before the `updateChangeStatus` call that writes `githubHeadSha`. The push succeeded and the published revision was known; it was simply not recorded.

Because this is the *re-promotion* path, it is not an edge case — it is the second and every subsequent promotion of any change. `githubHeadSha` is the record of which revision was actually published, so after a re-promotion it was stale (pointing at the previous head while the branch on GitHub had moved) or missing (rows predating the column). Anything reading that field to answer "what is currently published?" got a wrong answer, and nothing surfaced the discrepancy.

This writes the confirmed sha before the reuse path returns. **The reuse decision itself is untouched** — reusing a PR that matches the target is the correct answer, as the issue notes; only the write was missing.

### Scope held deliberately tight

- **Only the sha.** `buildUpdateChangeStatusStatement` skips `undefined` opts, so this assigns `status` and `github_head_sha` and nothing else. `promoted_at` / `promoted_by` are left alone — whether a re-promotion counts as a fresh promotion event is a product question this bug does not need answered, and I would rather it be a decision than a side effect.
- **Legacy rows.** When `publishedSha` is `undefined` (no `evaluatedSha`, so the staleness gate never ran) there is no confirmed revision and the write is skipped, matching the guard the create path already applies to its own `githubHeadSha` write.
- **No backfill** of rows already stale.

### One judgement call worth reviewing

**A failed write does not fail the promotion** — it is logged at `error` and the route still returns success. The branch is pushed and the PR exists, so the promotion genuinely happened; failing here would let a transient D1 error break a re-promotion that previously could not fail after the push, and the degraded outcome of a failed write is exactly today's behaviour, a stale sha. The create path stays strict because a failed write *there* loses the PR number itself and the next promotion would re-create the PR.

If you'd rather this path fail loudly, it is one branch to change.

### A pre-existing assertion had to go

`tests/changes.test.ts` asserted `expect(updateChangeStatus).not.toHaveBeenCalled()` on the re-promotion case. That assertion encoded the bug, so it is replaced rather than quietly relaxed — flagging it since a deleted assertion is the kind of thing worth a second look.

## Related issues

Closes #291

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## How was this verified?

Four cases on the re-promotion path, matching the regression the issue asks for — re-promotion with valid stored PR metadata after a force-push, asserting `githubHeadSha` is the new head rather than the old one:

1. **Records the revision it just published, not the one it replaced** — stored head `sha-first`, publishes `sha-second`, asserts the write carries `sha-second`, and asserts `promotedAt` is *not* in the payload.
2. **A legacy change with no `evaluatedSha` records no sha** — no gate ran, so nothing is written rather than something invented.
3. **Re-promotion still succeeds when the write fails** — covers the judgement call above.
4. The existing reuse case, retained, minus the assertion that encoded the bug.

Cases 1 and 3 **fail against the pre-fix route** — confirmed by stashing `src/routes/changes.ts` with the tests in place.

- [x] `npm test` — 2196 passed, 27 skipped, 0 failed (145 files)

## Checklist

- [x] Docs updated where the change makes them inaccurate — grepped `README.md`, `AGENTS.md`, `CONTRIBUTING.md` and `docs/` for `githubHeadSha` / `github_head_sha` / re-promotion semantics; no documentation describes this field. Nothing to correct.

---

## Summary by CodeRabbit

* **Improvements**
  * GitHub re-promotions now record the newly published workspace revision when evaluation is verified.
  * Promotions continue successfully even if recording the revision fails.
  * Legacy changes without an evaluated revision remain unchanged.

### DIFF
--- a/src/routes/changes.ts
+++ b/src/routes/changes.ts
@@ -1803,6 +1803,31 @@ app.post("/changes/:id/github-pr", async (c) => {
       prNumber: storedPr.number,
       repo: `${repo.owner}/${repo.repo}`,
     });
+    // The force-push above moved the PR's head, so the stored `githubHeadSha`
+    // now describes the PREVIOUS revision. This path is every re-promotion, so
+    // without this write the recorded published revision is stale (or missing,
+    // on rows predating the column) for every change promoted more than once,
+    // and nothing surfaces the discrepancy. Undefined only for a legacy change
+    // with no `evaluatedSha`, where the gate above never ran and there is no
+    // confirmed revision to record — the same guard the create path uses.
+    if (publishedSha !== undefined) {
+      const headShaResult = await updateChangeStatus(c.env.DB, logger, id, "promoted", {
+        githubHeadSha: publishedSha,
+      });
+      if (!headShaResult.success) {
+        // Not fatal: the branch is pushed and the PR exists, so the promotion
+        // genuinely happened. Failing here would let a transient D1 error break
+        // a re-promotion that previously could not fail after the push, and the
+        // degraded outcome is exactly the stale sha this write is fixing.
+        // Contrast the create path below, where a failed write loses the PR
+        // number itself and the next promotion would re-create the PR.
+        logger.error(
+          "Re-promotion succeeded but the published revision was not recorded",
+          headShaResult.error,
+          { changeId: id, publishedSha },
+        );
+      }
+    }
     return okOrFormRedirect(c, id, {
       changeId: id,
       github: {
--- a/tests/changes.test.ts
+++ b/tests/changes.test.ts
@@ -2953,19 +2953,20 @@ describe("POST /api/changes/:id/github-pr", () => {
     expect(updateChangeStatus).not.toHaveBeenCalled();
   });
 
+  const reusableStoredPr = {
+    status: "promoted" as const,
+    githubOwner: "acme",
+    githubRepo: "widgets",
+    githubBranch: "stratum/chg_abc123",
+    githubPrNumber: 7,
+    githubPrUrl: "https://github.com/acme/widgets/pull/7",
+    githubPrState: "open",
+  };
+
   it("re-promotion re-pushes the branch and reuses the existing PR", async () => {
     vi.mocked(getChange).mockResolvedValue({
       success: true,
-      data: {
-        ...acceptedChange,
-        status: "promoted",
-        githubOwner: "acme",
-        githubRepo: "widgets",
-        githubBranch: "stratum/chg_abc123",
-        githubPrNumber: 7,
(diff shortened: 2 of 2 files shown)


jlamoreaux:
@coderabbit review

PR #296, opened 2026-08-27
Recognise diff headers by position so a credential cannot dodge the secret scan

### LINKED ISSUE #280: Secret scanner skips any added line whose content begins with "++", so a credential can evade the always-on scan
## Summary

`SecretScanEvaluator` decides which diff lines to inspect with:

```ts
if (!line.startsWith("+") || line.startsWith("+++")) return;
```

The second test is meant to skip the `+++ b/path` file header. But a diff prefixes every added line with `+`, so a **file line whose own content starts with `++`** arrives as `+++…` and is skipped along with the headers. A credential on such a line is never scanned.

This is pre-existing and is not introduced by any open PR. Found while pre-auditing #232, which feeds the scanner a synthetic diff — that construction turns out to be correct and consistent with real diffs, and it inherits this behaviour rather than causing it.

## Reproduction

Verified by calling the evaluator directly (`passed = true` means nothing was detected):

| Input | `passed` |
| --- | --- |
| `+++ b/a.txt` / `+AKIAIOSFODNN7EXAMPLE` | **false** — detected, correct |
| `+++ b/a.txt` / `+++AKIAIOSFODNN7EXAMPLE` | **true** — missed |
| a full unified diff (`diff --git` / `---` / `+++` / `@@`) whose added line is `++AKIAIOSFODNN7EXAMPLE` | **true** — missed |

The third row is the important one: the evasion is not an artefact of a synthetic diff. It works against the real diff the scan normally runs on, so it applies to the always-on blocking scan every change is subject to.

## Why the obvious one-character fix is not enough

Tightening the header test to `line.startsWith("+++ ")` (with a trailing space) fixes the exact case above, because git always writes `+++ b/path`. It does not fix content beginning with `++ ` (two pluses and a space), which still renders as `+++ ` and is still skipped.

The robust fix is to parse the diff structurally rather than by line prefix: track position, treat `---`/`+++` as headers only where a header is expected (immediately after `diff --git`, before the first `@@`), and treat everything after a hunk header as content. Then strip exactly one `+` and scan the remainder, so the scanned text is the real file line.

## Severity

Defense-in-depth evasion rather than privilege escalation: it requires someone deliberately shaping a line to dodge the scanner, and the scan is one control among several. But it is a blocking, always-on control that can be stepped around with knowledge of the implementation, and the fix is contained to one function.

Worth adding regression cases for both `++`-prefixed and `++ `-prefixed content when this is fixed, since neither is covered today.

jlamoreaux:
## Summary

`SecretScanEvaluator` decided which diff lines to inspect by prefix. `main` already carries the one-character tightening the issue anticipated — `startsWith("+++ ")`, with the trailing space — which closes `++`-prefixed content. It does **not** close the case the issue explicitly says it will not close: a file line whose own content begins with `++ ` (two pluses and a space) arrives in the diff as `+++ …`, is read as a file header, and is never scanned.

This is a blocking, always-on control every change passes through, and it could be stepped around by anyone who knows the implementation.

The fix is the structural parse the issue asks for. A single-pass walk holds two pieces of state — whether we are inside a hunk body, and whether the previous line was a `--- ` old-file header — and classifies from git's own structure:

- `diff --git ` starts a file and resets both flags.
- `@@` starts a hunk body.
- Before any hunk, `--- ` is the old-file header.
- `+++ ` is the new-file header **only when it immediately follows that `--- ` line**. Git always emits the pair adjacently, so a `+++ …` line anywhere else is content.
- Everything else beginning with `+` is an added line: strip exactly one `+`, scan the rest — so the patterns see the real file line rather than one with a marker welded to its first token.

There is now no shape a content line can take to be mistaken for a header.

## Two deliberate deviations, both worth review

**1. Anchored on the header pair, not on `@@`.** The issue says to "treat everything after a hunk header as content", which is the stricter reading. But the suite's own `makeDiff` helper emits `diff --git` / `--- ` / `+++ ` and then added lines with **no `@@` at all** — a parser gated on `@@` would skip every added line in the existing corpus. A scanner that silently scans nothing is a far worse failure than the one being fixed, so the parse is tolerant of a hunk-less diff (and of a bare list of `+` lines) while still classifying headers correctly.

**2. An unpaired `+++ ` line is now treated as content.** A `+++ ` with no `--- ` before it is not where git puts a header. Erring toward scanning is the right direction for a blocking security gate: a false positive is a blocked change with a stated reason, a false negative is a leaked credential. A real diff always carries the pair, so a genuine header is never caught by this.

That second point **changed an existing test.** `still skips the real +++ b/path file header` asserted that a bare `+++ b/<secret-shaped-path>.ts` fragment is skipped. It is rewritten to use a realistic `diff --git` / `--- ` / `+++ ` / `@@` sequence, where the intent it was protecting — a header path containing a secret-shaped substring must not be reported as content — still holds and is still asserted. Flagging it since a rewritten security assertion deserves a second pair of eyes.

### Explicitly not in scope

- New patterns, threshold changes, entropy tuning.
- **Changing the reported line number** from a diff-line index to a file-line number. The structural walk makes real file numbers derivable, but changing what the number means is behaviour-visible and existing consumers read the current one.
- Removed and context lines — the gate is about what a change *introduces*.

## Related issues

Closes #280

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## How was this verified?

Five cases in the `added lines whose own text starts with ++` block, covering both shapes the issue names as uncovered:

1. **`++`-prefixed content** (`+++text`) — already passing on `main`; retained as a guard.
2. **`++ `-prefixed content** (`+++ text`) — the case the trailing-space fix does not close.
3. **An unpaired `+++ ` line** — documents deviation 2 above.
4. **A multi-file diff** where the second file hides the credential — guards against header state leaking across files.
5. **A removed line starting with `-- `** (a SQL comment rendering as `--- …` inside a hunk) — guards against a removal setting up header state that then swallows the next added line.

Cases 2–5 **fail against the pre-fix scanner** and pass with it. I initially had cases 4 and 5 built with the wrong marker depth — they passed in both directions, which is how I caught it; the fixtures are corrected so both genuinely exercise the evasion.

- [x] `npm test` — 2197 passed, 27 skipped, 0 failed (145 files)

## Checklist

- [x] Docs updated where the change makes them inaccurate — swept `README.md`, `AGENTS.md`, `CONTRIBUTING.md`, `docs/`, and `website/`. The markdown descriptions ("a secret scan is always on and always blocking") stay true. The one piece of prose this invalidates is the `scanContentForSecrets` JSDoc, which described the skipped-as-a-header behaviour as current; it is corrected in this diff.
- [x] No secrets, tokens, or personal data committed — new fixtures follow the file's existing convention of splitting every credential-shaped literal across at least two parts

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved secret scanning in diffs to detect secrets in added lines beginning with `++`.
  * More accurately distinguishes file headers from content, including across multiple files.
  * Preserved correct handling of removed lines and diff boundaries.

* **Documentation**
  * Clarified how direct content scanning handles lines beginning with `++`.

### DIFF
--- a/src/evaluation/secret-scanner.ts
+++ b/src/evaluation/secret-scanner.ts
@@ -149,12 +149,13 @@ function matchSecret(line: string): string | undefined {
 /**
  * Scan raw file content — not a diff — for secrets.
  *
- * Callers that hold the literal bytes about to be committed must use this
+ * Callers that hold the literal bytes about to be committed should use this
  * rather than synthesising a diff to feed `SecretScanEvaluator.evaluate`.
- * Prefixing content lines with "+" is not a lossless encoding: a content line
- * that itself starts with "++" becomes "+++…", which the diff reader skips as
- * a file header, and the secret on it is never scanned. Scanning the content
- * directly removes that escape entirely.
+ * Prefixing content lines with "+" is not a lossless encoding — a content line
+ * starting with "++" is indistinguishable from a "+++" file header by prefix
+ * alone — so the evaluator recovers the distinction from the diff's structure
+ * instead. Scanning content directly needs no such recovery, and is the
+ * simpler thing to reason about when the bytes are already in hand.
  *
  * @returns One issue string per finding, empty when the content is clean.
  */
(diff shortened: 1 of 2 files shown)


jlamoreaux:
@coderabbit review

jlamoreaux:
@coderabbit review

PR #298, opened 2026-08-27
Persist the import clone depth so retry and re-import stop shallowing a full clone

### LINKED ISSUE #279: Import clone depth is accepted but never persisted, so retry and sync silently fall back to depth 10
## Summary

PR #225 adds `validateCloneDepth` and lets a caller choose a clone depth (`1..1000`, or `0`/`"full"` for full history) when importing. The depth reaches the first import job, but it is never stored, so every later job for that project re-derives a hardcoded `depth: 10`.

Found while pre-auditing #225 against current `main`. Not a blocker for that PR's stated scope — filing separately rather than widening it.

## What happens

`import_jobs` (migrations/010_import_jobs.sql) has columns for `source_url`, `branch`, status and progress, but **no `depth` column**, and #225 adds no migration. So:

1. **Retry / re-queue** (`src/routes/projects.ts`, the `queueImportJob` call in the re-import handler). It reconstructs the job from the stored row:

   ```ts
   const githubUrl = existing.sourceUrl;
   const branch = existing.branch ?? "main";
   ...
   await queueImportJob(c.env.IMPORT_QUEUE, { ..., githubUrl, branch, depth: 10 });
   ```

   `sourceUrl` and `branch` survive because they are columns. Depth is not, so it is re-hardcoded. A user who imported with `"full"` and hit a transient failure gets a depth-10 retry, with no error and no indication the depth changed — the opposite of what they asked for, silently.

2. **Sync** (`queueSyncJob`, added by #215) likewise passes `depth: 10`.

Case 2 is only visible **after merging `main` into #225**. The two changes touch different regions of `projects.ts`, so git reports no conflict and both branches are green on their own; the gap exists only in the merged result. Reproduced by merging locally and grepping the result — two hardcoded `depth: 10` call sites remain alongside the validated one.

## Why it matters

Depth is the difference between a usable clone and a broken one for the workflows that need history — `git log`, `git blame`, bisect, and the backup path that explicitly needs full history. Falling back to 10 turns a deliberate "full" into a shallow clone at the first retry or sync, and nothing surfaces the change.

## Suggested direction

Persist it:

- add `depth INTEGER` to `import_jobs` (nullable; `NULL` = never specified, distinct from `0` = full history — the sentinel `0` is meaningful, so a `NOT NULL DEFAULT 0` column would be wrong);
- write it when the job is created, read it on the retry and sync paths instead of the literal;
- keep `DEFAULT_CLONE_DEPTH` as the fallback only when the column is `NULL`.

Note for whoever picks this up: `migrations/` currently has four open PRs holding 036–039 (#221, #222, #228, #234), so this would be 040.

## Scope note

Diagnosis-quality and data-fidelity, not correctness or security — the import still succeeds, just shallower than requested.

jlamoreaux:
## Summary

`validateCloneDepth` lets a caller choose a clone depth (`1..1000`, or `0`/`"full"` for full history). The depth reached the first import job and was then thrown away — `import_jobs` had no `depth` column, so every later job for the project re-derived a hardcoded literal. A user who imported with `"full"` and hit a transient failure got a depth-10 retry: the opposite of what they asked for, with no error and no indication the depth had changed.

Depth is the difference between a usable clone and a broken one for anything needing history — `git log`, `git blame`, bisect, and the backup path that explicitly needs full history.

### The migration

`040_import_jobs_depth.sql` adds `depth INTEGER` to `import_jobs`. **Nullable with no default, and that is load-bearing rather than stylistic**: `0` means full history, so `NULL` (never specified → fall back to `DEFAULT_CLONE_DEPTH`) has to stay distinguishable from it. A `NOT NULL DEFAULT 0` column would read every pre-existing row as a full-history request — exactly the sentinel trap the issue flags.

036–039 are merged, so 040 is free, as the issue predicted.

### Reading it back

`createImportJob` writes it; `rowToImportJob` maps it **only when non-null**, so the default is applied by consumers rather than erased at the storage layer.

Every path that derives a later job from a stored one now reads it:

| Path | Source of depth |
|---|---|
| Retry (queued) | `existing.depth ?? DEFAULT_CLONE_DEPTH` |
| Retry (queue-less, direct) | `existing.depth ?? DEFAULT_CLONE_DEPTH` |
| Form re-submit (queued) | `existingImport.data.depth ?? DEFAULT_CLONE_DEPTH` |
| Form re-submit (queue-less, direct) | `existingImport.data.depth ?? DEFAULT_CLONE_DEPTH` |
| Sync route (queued) | `getLatestImportDepth() ?? DEFAULT_CLONE_DEPTH` |
| Sync route (queue-less fallback) | passed through as a **required** `processSyncJob` parameter |
| GitHub push webhook | `getLatestImportDepth() ?? DEFAULT_CLONE_DEPTH` |

The sync entry points create a *new* job, so they inherit via a new `getLatestImportDepth()` — a narrow single-column read rather than `getImportProgress`, which returns the whole row and parses the `logs`/`errors` JSON blobs the webhook path has no use for.

One trap worth calling out, since getting it wrong is silent: `getLatestImportDepth` is called **before** the new job is created. That job would otherwise become the "most recent" one and the lookup would read back the value it is trying to inherit. It is commented at both call sites.

`processSyncJob` takes `depth` as a **required** parameter rather than a defaulted one: a default there is indistinguishable from a caller that forgot, which is exactly how that path kept re-importing at 10.

### Two corrections to earlier revisions of this description

Flagging both rather than quietly editing, since each changed what the PR claims.

**1. The sync-path win is narrower than first stated.** `syncOrImportProject` routes a project with a parseable Artifacts remote — i.e. every established project — to an **incremental fetch that ignores this depth** and uses its own `SYNC_FETCH_DEPTH` window. The depth reaches only the **fallback full import**, taken when a project has no Artifacts remote because its initial import never completed. That routing is documented in `src/services/project-sync.ts` and predates this PR (untouched by #276). Carrying the recorded depth there is still right — the fallback re-imports the whole project, so a hardcoded 10 is precisely where a wrong depth costs history — it is just not the common path. The title was narrowed from "retry and sync" to "retry and re-import" to match.

**2. My audit claim was false.** An earlier revision said *"no hardcoded depth literal remains outside `validation.ts`'s own default — verified by grep."* That grep matched `depth: 10` with a colon, which finds neither an omitted argument nor an assignment with `=`, and it missed two real sites that CodeRabbit's review then caught (see the discussion below). Both are fixed in `20d2d43`. The audit was redone by enumerating every `processImportJob` / `processSyncJob` / `syncOrImportProject` call site rather than grepping for literals.

### Upgrade and deploy safety

Existing rows are `NULL` and fall back exactly as today, so **no project changes behaviour on upgrade**. `ALTER TABLE ADD COLUMN` is additive: old code ignores the column, new code tolerates `NULL`, so deploy order does not matter.

### Explicitly not in scope

- **The daily cron** (`syncAllProjects`, `src/routes/sync.ts`) passes no depth, so its legacy fallback import takes `importFromGitHub`'s own default. Same defect class, but six assertions in `tests/sync-cron.test.ts` encode "the cron passes no depth" as the current contract. Raised [in a comment](https://github.com/stratum-eng/stratum/pull/298#issuecomment-5439699851) with a ready patch rather than changed unilaterally — your call.
- **Making the incremental sync honour a recorded depth.** `SYNC_FETCH_DEPTH` plus the deepening loop is a deliberate design that #276 has just reworked.
- Letting a user *change* the depth at retry or sync time. This restores the depth they already chose.
- Backfilling `depth` on existing rows.
- `bulk-import.ts`, which has no depth to record and so stores `NULL` and falls back.

## Related issues

Closes #279

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## How was this verified?

Ran the CI gate order locally (lint → typecheck → test), against the branch **with `origin/main` merged in** (#276 landed after this PR opened; it touches `src/storage/git-ops.ts` and its sync tests, none of which this PR touches, so the merge is clean).

Twelve new cases, all confirmed to fail against the pre-fix code by reverting the source with the tests in place:

**Storage round trip** (`tests/import-clone-depth.test.ts`) — against `makeSqliteD1`, a real SQLite engine with every migration applied, so migration 040 and the SQL reading it are genuinely executed rather than stubbed:
- a shallow depth round-trips;
- **`0` survives as distinct from unset**, asserted through `depth ?? DEFAULT_CLONE_DEPTH` itself — the exact expression that would eat it;
- a caller with no depth records `NULL` and consumers fall back;
- a row forced to `NULL` (simulating pre-migration) reads as unset;
- the *most recent* job's depth wins, with `started_at` nudged so ordering is not decided by insertion luck;
- a project with no jobs reads as unset.

**Retry and re-trigger** (`tests/import-fidelity.test.ts`) — the issue's headline case: a full-history import that fails and is retried is re-queued with `depth: 0`, asserted both positively and as *not* `DEFAULT_CLONE_DEPTH`; a shallow-but-deep (500) case; and the queue-less re-trigger importing at a recorded `0` rather than `10`.

**Queue-less sync fallback** (`tests/incremental-sync-wiring.test.ts`) — the fallback import receives its caller's depth rather than a hardcoded 10.

**Webhook sync path** (`tests/webhook-push.test.ts`) — a recorded depth reaches both `queueSyncJob` and `createImportJob`; `0` is preserved rather than defaulted; and no recorded depth falls back to the default.

Two pre-existing webhook tests failed when I first wired this up, because that file's `../src/storage/imports` module mock did not expose the new function — the stub is updated, not the assertions.

- [x] `npm test` — 2214 passed, 27 skipped, 0 failed (146 files)

## Checklist

- [x] Docs updated where the change makes them inaccurate — swept `docs/user-guide/importing.md`, `website/src/content/docs/guides/importing.md`, `README.md` and `AGENTS.md`. The `depth` option table stays accurate, and nothing documented claimed depth was *or* was not retained across retry — the docs are silent, not wrong, so nothing needed correcting.

## Summary by CodeRabbit

* **New Features**
  * Import jobs now retain their configured clone depth, including full-history imports.
  * Webhook-triggered syncs, retries, and re-triggered imports reuse the previously recorded clone depth.
  * Imports without a recorded depth continue using the default setting.

* **Bug Fixes**
  * Prevented imports from unexpectedly reverting to the default or fixed depth during retries and webhook syncs.
  * Preserved explicit full-history (`0`) and custom-depth configurations.

### DIFF
--- a/migrations/040_import_jobs_depth.sql
+++ b/migrations/040_import_jobs_depth.sql
@@ -1,0 +1,13 @@
+-- Persist the clone depth an import job ran under (#279).
+--
+-- Without this column, retry, re-queue and both sync paths rebuild the job
+-- from the stored row and re-derive a hardcoded DEFAULT_CLONE_DEPTH, silently
+-- turning a deliberate full-history import into a shallow one at the first
+-- transient failure.
+--
+-- Nullable with no default, deliberately: 0 is a MEANINGFUL value meaning
+-- "full history", so NULL (never specified -> fall back to
+-- DEFAULT_CLONE_DEPTH) has to stay distinguishable from it. A
+-- `NOT NULL DEFAULT 0` column would read every pre-existing row as a
+-- full-history request.
+ALTER TABLE import_jobs ADD COLUMN depth INTEGER;
--- a/src/github/webhooks.ts
+++ b/src/github/webhooks.ts
@@ -125,10 +125,15 @@ async function handlePush(
 
   const importId = crypto.randomUUID();
   const sourceUrl = project.sourceUrl ?? project.remote;
-  // The depth the project was actually imported at. Read before the enqueue
-  // (and before the job below is created, which would otherwise become the
-  // "most recent" job this reads back) so a webhook-driven sync cannot quietly
-  // shallow a full-history project.
+  // The depth the project was actually imported at, replacing a hardcoded 10.
+  // Read before the enqueue (and before the job below is created, which would
+  // otherwise become the "most recent" job this reads back).
+  //
+  // Reaches the FALLBACK FULL IMPORT only — the branch `syncOrImportProject`
+  // takes for a project with no parseable Artifacts remote. The incremental
+  // fetch every established project takes ignores this and uses its own
+  // SYNC_FETCH_DEPTH window. Still worth carrying: the fallback re-imports the
+  // whole project, so a wrong depth there is exactly where it is costly.
   const syncDepth =
     (await getLatestImportDepth(env.DB, project.namespace, project.slug, logger)) ??
     DEFAULT_CLONE_DEPTH;
--- a/src/github/webhooks.ts
+++ b/src/github/webhooks.ts
@@ -6,11 +6,12 @@
 import { Hono } from "hono";
 import { createChange, getChangeByGitHubBranch, updateChangeStatus } from "../storage/changes";
 import { getProjectByGitHubRepo } from "../storage/github-bridge";
-import { createImportJob } from "../storage/imports";
+import { createImportJob, getLatestImportDepth } from "../storage/imports";
 import { getWorkspace } from "../storage/state";
 import { getSyncStatus, setSyncInProgress } from "../storage/sync";
 import { type Env, projectDefaultBranch } from "../types";
 import { type Logger, createLogger } from "../utils/logger";
+import { DEFAULT_CLONE_DEPTH } from "../utils/validation";
 
 const app = new Hono<{ Bindings: Env }>();
 
@@ -124,6 +125,13 @@ async function handlePush(
 
   const importId = crypto.randomUUID();
   const sourceUrl = project.sourceUrl ?? project.remote;
+  // The depth the project was actually imported at. Read before the enqueue
+  // (and before the job below is created, which would otherwise become the
+  // "most recent" job this reads back) so a webhook-driven sync cannot quietly
+  // shallow a full-history project.
+  const syncDepth =
+    (await getLatestImportDepth(env.DB, project.namespace, project.slug, logger)) ??
+    DEFAULT_CLONE_DEPTH;
 
   // Enqueue FIRST — only write state flags after a successful send().
   const { queueSyncJob } = await import("../queue/import-queue");
@@ -134,7 +142,7 @@ async function handlePush(
     slug: project.slug,
     githubUrl: sourceUrl,
     branch: sourceBranch,
-    depth: 10,
+    depth: syncDepth,
     trigger: "webhook",
   });
 
@@ -149,6 +157,7 @@ async function handlePush(
       slug: project.slug,
       sourceUrl,
       branch: sourceBranch,
+      depth: syncDepth,
     },
     logger,
   );
--- a/src/routes/projects.ts
+++ b/src/routes/projects.ts
@@ -22,6 +22,7 @@ import {
   createImportJob,
   deleteImportJob,
   getImportProgress,
+  getLatestImportDepth,
   isImportCancelled,
   recoverStalledImport,
   updateImportStatus,
@@ -450,7 +451,9 @@ app.post(
                   slug,
                   githubUrl: existingImport.data.sourceUrl,
                   branch: existingImport.data.branch ?? "main",
-                  depth: DEFAULT_CLONE_DEPTH,
+                  // The depth the job was created with, not a fresh literal —
+                  // a re-trigger must not quietly shallow a full-history import.
+                  depth: existingImport.data.depth ?? DEFAULT_CLONE_DEPTH,
                   initiatedBy: userId,
                 });
               } catch (queueError) {
@@ -564,6 +567,7 @@ app.post(
           slug,
           sourceUrl: body.url,
           branch,
+          depth,
         },
         logger,
       );
@@ -1442,7 +1446,7 @@ app.post(
           slug,
           githubUrl,
           branch,
-          depth: DEFAULT_CLONE_DEPTH,
+          depth: existing.depth ?? DEFAULT_CLONE_DEPTH,
           // Without this the retry path is the one case that never emails the
           // person waiting on the import — which is the gap this PR closes.
           initiatedBy: userId,
@@ -1477,7 +1481,7 @@ app.post(
           githubUrl,
           branch,
           logger,
-          DEFAULT_CLONE_DEPTH,
+          existing.depth ?? DEFAULT_CLONE_DEPTH,
         ).catch((error) => {
           logger.error(
             "Unhandled error in background import retry",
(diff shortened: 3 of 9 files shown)


jlamoreaux:
The merge-risk finding was correct, and both paths are now fixed in `20d2d43`.

**What was missed and why.** The queued paths inherited the recorded depth, but two fallbacks did not:

1. The form re-submit handler's queue-less branch called `processImportJob` with six arguments, **omitting `depth` entirely** and silently taking that parameter's own `DEFAULT_CLONE_DEPTH`. I had fixed the queued branch of that same handler but not its `else`.
2. `processSyncJob` (the queue-less sync fallback) carried `const depth = 10; // Default depth for the legacy import fallback`.

Both were invisible to the audit I claimed in the PR description. **That claim — "no hardcoded depth literal remains outside `validation.ts`'s own default — verified by grep" — was wrong**, and I've corrected it in the description. My grep matched `depth: 10` with a colon, which finds neither an omitted argument nor an assignment with `=`. I re-audited by enumerating every `processImportJob` / `processSyncJob` / `syncOrImportProject` call site instead of grepping for literals.

`processSyncJob` now takes `depth` as a **required** parameter rather than a defaulted one — a default there is indistinguishable from a caller that forgot, which is exactly how the path kept re-importing at 10. Its only production caller already had `syncDepth` in scope.

Two regression tests, both confirmed to fail against the pre-fix route: the queue-less re-trigger imports at a recorded depth of `0` rather than `10`, and the fallback sync carries its caller's depth into the import.

---

### One more site, deliberately not fixed here — your call

The re-audit turned up a third instance that the review did not flag: **the daily cron**, `syncAllProjects` in `src/routes/sync.ts`, passes no `depth` to `syncOrImportProject` at all, so its legacy fallback import takes `importFromGitHub`'s own `depth = 10`. A full-history project re-imported by the cron is shallowed the same way.

I'm not pushing that fix, because it isn't equivalent to the two above: `tests/sync-cron.test.ts` asserts `undefined` for that argument in **six** places, with the comment *"The cron passes no depth; the shared helper forwards it as undefined, so importFromGitHub falls back to its own default."* That's a documented contract at a separate entry point, and changing it means rewriting those assertions. Happy to do it — just not unilaterally.

The patch, if you want it:

```ts
// src/routes/sync.ts, inside syncAllProjects' per-project loop
const inheritedDepth = await getLatestImportDepth(
  env.DB,
  project.namespace,
  project.slug,
  projectLogger,
);
const outcome = await syncOrImportProject(
  env.ARTIFACTS,
  {
    remote: project.remote,
    artifactsRepoName: artifactsRepoName(project),
    sourceUrl,
    branch,
    depth: inheritedDepth ?? DEFAULT_CLONE_DEPTH,
    logContext: { namespace: project.namespace, slug: project.slug },
  },
  projectLogger,
);
```

Behaviour is unchanged for any project with no recorded depth (`undefined` → `importFromGitHub`'s 10 becomes an explicit 10); it only differs once a depth is stored. The six `undefined` assertions in `tests/sync-cron.test.ts` become `DEFAULT_CLONE_DEPTH`. I verified this compiles and that those six are the only failures it causes.

Say the word and I'll push it here, or open it as a follow-up — whichever you prefer.

---

### On the Docstring Coverage check

It reports 25% against an 80% threshold. I'd rather not add docstrings purely to clear it: `AGENTS.md` says *"Comments explain **why**, not **what**"*, and the functions it's counting are mostly test helpers and small local closures where a docstring would restate the signature. The two genuinely public additions in this diff — `getLatestImportDepth` and the `depth` field on `ImportProgress` — already carry JSDoc explaining the `NULL`-vs-`0` distinction, which is the non-obvious part. Happy to be overruled if the threshold is a hard gate.

---

PR #319, opened 2026-08-28
Multi-branch support, webhook project_id scoping, and a base commit in the evaluator payload (#181, #235, #274)

### LINKED ISSUE #181: No multi-branch support: server-side git ops hardcode `ref: "main"`
## Concern

Stratum is effectively a single-branch forge. Every server-side git operation targets `main` literally.

- `src/storage/git-ops.ts` hardcodes `ref: "main"` / `refs/heads/main` in ~26 places (clone, fetch, push, merge, resolveRef, checkout).
- There is no branch CRUD and no ref-scoped browsing: no `/branches` route, no `?ref=` param on file/log/tree endpoints, no branch switcher in the UI (`src/routes/ui.tsx`).
- "Branching" is only modeled as fork-per-workspace (each workspace is a separate Artifacts repo with its own `main`).

### Default-branch mismatch hazard
The push gate resolves the real default branch (`src/routes/git-http.ts:601` uses `project.sourceDefaultBranch || githubDefaultBranch || "main"`), but `cloneRepo`/`pushMain`/`revertToCommit` never consult it. A repo whose default is `master`/`trunk` (persisted at `src/routes/projects.ts:574`) has every internal git op (browse, diff, merge, backup, snapshot) attempt `ref: "main"` and fail.

## Impact
No engineering team can operate without branches. This is the single largest gap vs GitHub.

## Suggested direction
Introduce a ref abstraction plumbed through clone/browse/diff/merge; add branch list/create/delete endpoints and a ref selector in the browser; make every internal op consult the project's resolved default branch.

### LINKED ISSUE #235: Backfill webhooks.project_id and remove the name-based tenant fallback

### LINKED ISSUE #274: Webhook payload carries no base commit, so a receiver can evaluate the wrong tree
The webhook evaluator posts `{ diff, policy }` (`src/evaluation/webhook-evaluator.ts:52`) and nothing else. The diff is generated against the project's `main` at evaluation time (`listFilesAtCommit(baseFs, "main", ...)` / `readFileAtCommit(baseFs, "main", ...)` in `src/storage/git-ops.ts`), but the commit that `main` pointed at is not part of the request.

A receiver therefore has no way to reproduce the base. The natural implementation — keep a mirror, check out `main`, `git apply` the diff — evaluates against whatever `main` happens to be when the request lands.

## Failure modes

- **Wrong verdict.** `main` advances between diff generation and delivery. `git apply` succeeds against the newer tree because the hunks still have context, and the suite runs against a combination that no change actually proposes. A pass or fail here does not describe the change under evaluation.
- **Spurious failure.** The same race with a conflicting hunk fails `git apply`, and the receiver reports the change as broken when it is fine.
- **Silent.** Neither case is distinguishable from a real result at either end.

`change-flow.ts` already resolves and persists a base sha for the change (`const baseSha = await resolveProjectHead(...)`), so the value exists — it just is not forwarded to the evaluator.

## Suggested direction

Add the base commit to the request body and document it as part of the contract, so a receiver can check out that exact commit or reject the request when it cannot. Adding a field is backward compatible for receivers that ignore it.

Worth a regression that advances `main` between diff generation and delivery and asserts the request is not evaluated against the newer commit.

## Provenance

Found by CodeRabbit while reviewing #217 (the bring-your-own-CI guide). #217 documents the contract as it ships today and links here for the caveat; it does not change the payload.

jlamoreaux:
## Summary

Three open issues, one commit each (plus a fourth carrying review fixes). They are independent — the commits are separable if you would rather land them apart.

### #181 — multi-branch support

PR #226 landed the default-branch hazard and explicitly left "branch CRUD / ref-scoped browsing / branch switcher" open. This closes that remainder.

**The premise had to be corrected first.** No path in the codebase puts a second `refs/heads/*` on a project repo: `importFromGitHub` imports a single branch, `restoreProjectRepo` pushed only the default branch and tags, every merge and sync targets the default branch, and the push gate refuses everything else. So this is not "the UI fails to show what `git clone` already shows" — there was nothing to show. **Branch creation is the enabling feature**; listing and ref-scoped browsing are what make it usable.

**Listing reads the ref advertisement, not a clone.** The obvious design — model `collectRepoBranches` on `collectRepoTags` and fetch each branch — does not work: isomorphic-git translates fetched head refs through the remote's refspec into `refs/remotes/origin/*`, so a clone's `refs/heads/` holds exactly one branch no matter how many were fetched. `git.getRemoteInfo` already carries every branch name and tip oid in two subrequests, which is all a listing needs. That also keeps `?ref=` validation off the clone path, where it would otherwise have cost ~400 subrequests per read against a ~1000 cap.

**Branch creation cannot introduce content.** The start point must resolve to an object the repository already holds — the default tip, another branch's tip, or a full 40-hex sha verified as a commit in a full-history clone. Tags are refused as start points: a tag can point at a blob, an annotated tag's peeled commit is advertised by upload-pack but not receive-pack (so `git.push` walks the local graph and fails outside the shallow window), and a tag may sit outside both the default branch's history and the `MAX_TAGS`-capped fetch a backup makes. This invariant is also what lets backup record refs only and leave `walkRepoObjects` untouched.

**Enforcement is one application-level check**, in `createBranchRef`. The push goes straight to the Artifacts remote with a minted write token and bypasses `src/routes/git-http.ts` entirely. That is deliberate and stated plainly in the code so no reader infers a defense in depth that is not there. The push gate itself is unchanged — pushing *content* to a non-default project ref over git stays refused.

Also included: `GET/POST/DELETE /api/projects/:ns/:slug/branches`, `?ref=` on `files`/`content`/`log` and on the UI repo and blob views, a `/branches` page, a no-JS `<form method="get">` branch switcher, and branch refs in backup and restore.

Explicit non-goals: changes or merges targeting a non-default branch, branch protection, branch rename, cross-branch compare, and `?ref=` accepting a tag or raw sha (a name that is both a branch and a tag is refused as ambiguous rather than resolved — `refpaths` checks `refs/tags/` *before* `refs/heads/`, so resolving it would serve the wrong tree under a URL that says otherwise).

### #235 — webhook tenant scoping, step 3

Steps 1–2 (the admin backfill and its `remainingNullRows` verification) landed in #268 and #287. `webhookBelongsToProject` and `listWebhooks` still matched `project_id = ? OR (project_id IS NULL AND project = ?)`, so a pre-migration-025 row could be read, mutated, and delivered to by a same-named project in another namespace.

Rather than delete the clause and leave the shape that produced it, `listWebhooks` now takes the project **id** in place of the name, so a name-scoped lookup is not expressible at the call site; `createWebhook`'s `projectId` is required for the same reason, so no new NULL row can undo the backfill. An unstamped legacy row belongs to no project — invisible to management, delivered to by nothing. The queue consumer drops an outbox row carrying no `project_id` with a WARN rather than resolving it by name.

### #274 — base commit in the webhook evaluator payload

The evaluator posted `{diff, policy}`, so a receiver could only apply the hunks to whatever the default branch was when the request landed. `getDiffBetweenRepos` already resolved the base tip — it needs it, because `readBlob` will not accept a ref name as an oid — and discarded it. It is now returned as `baseOid` and reaches the evaluator through a new optional `EvaluationContext`.

Deliberately the base from the diff's own clone at all three call sites: change creation resolves `baseSha` *before* the diff clone, and the re-evaluate route's `change.baseSha` records the base at creation, which is exactly the value that has gone stale by re-evaluation.

### Review round

Two adversarial reviews ran against the first three commits. The fourth commit fixes what they found. The one worth calling out: **restore authenticated its branch push as GitHub**, sending an Artifacts token verbatim instead of stripping its `?expires=` suffix — every restore carrying a branch would have failed, after main and the tags had already landed. It was invisible to the suite, which mocks `git.push` and never asserted the credential.

Also fixed: restore guards weaker than the request path they mirror (a non-commit tip, a ref-path collision aborting the whole restore, `HEAD` accepted as a branch name); backup silently recording fewer branches than the repo has; `DELETE` failing to find the name at all when the namespace arrived percent-encoded (Hono decodes params and paths differently, and this repo's own UI builds URLs the failing way); and four opaque 500/502s that are now named 4xx.

Four issues are **accepted rather than fixed**, recorded in `.claude/ship/TASKS.md`. The load-bearing one: the create pre-check is not a lock. `git.push` accepts a descendant oid without `force`, so two concurrent creates of the same name can let the second fast-forward the first. Closing it needs a compare-and-swap the transport does not offer, and the git-http gate already excludes every other writer of these refs.

## Related issues

Closes #181
Closes #235
Closes #274

## Type of change

- [x] Bug fix
- [x] New feature
- [x] Documentation

## How was this verified?

- [x] `npm run typecheck`
- [x] `npm test` — 2345 passed, 27 env-gated skips (unchanged from main), 0 failures
- [x] `npm run lint`

~90 new tests. The branch mechanics are tested against **real in-memory repos** rather than a wholly-mocked isomorphic-git, deliberately: the `refs/remotes` translation, `push({delete})`'s local-ref requirement, and the tag-before-branch `refpaths` order are all invisible to a mocked suite, and all three sank the first design. Three bugs of my own surfaced that way during implementation — pushing from a shallow clone after verifying the start point in a full-history one, branching from an annotated tag onto the tag object, and reporting a transport failure as a name conflict.

## Checklist

- [x] Docs updated where the change makes them inaccurate — `openapi.yml`, the projects endpoint reference, and the BYO-CI contract in `docs/user-guide/ci-integration.md`
- [x] The web UI change stays server-rendered with no client-side JavaScript — the branch switcher is a GET form

## Noted separately, not fixed here

PR #226's description claims it threaded the resolved default branch through "all ~26" hardcoded `ref: "main"` sites. `batchMergeWorkspaces` and `mergeStagedCommits` in `src/storage/git-ops.ts` still hardcode `main` (≈ lines 1204, 1229, 1275, 1284, 1399, 1406). That is a live bug for any project imported with a `master`/`trunk` default — batch merge targets a ref the repo does not have. It is a merge-path bug rather than multi-branch support, so it wants its own change.

https://claude.ai/code/session_01P11maBt5wGR783noPFLpqs

---
_Generated by [Claude Code](https://claude.ai/code/session_01P11maBt5wGR783noPFLpqs)_

## Summary by CodeRabbit

- **New Features**
  - Added project branch listing, creation, and deletion.
  - Browse repositories, files, content, and commit history by branch with branch switching.
  - Repository backups now preserve eligible branches during snapshot and restore.
  - CI evaluation payloads include the exact source commit when available.

- **Bug Fixes**
  - Added clear handling for invalid, unknown, or ambiguous branch references.
  - Webhook delivery now uses canonical project identifiers, preventing cross-project or ambiguous deliveries.

- **Documentation**
  - Updated API and CI integration documentation for branch support and source-commit handling.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Stratum issues opaque bearer tokens and has no `token_endpoint` or `jwks_uri`, and RFC 9728
   metadata is resolved from the API origin, which these docs do not serve. Advertising either
   would send agents to a URL that cannot answer them.
+- **Multi-branch support (#181).** Create, list and delete branches on a project
+  (`GET/POST/DELETE /api/projects/:ns/:slug/branches`), browse any of them with
+  `?ref=` on the files, content and log endpoints and in the web UI, and switch
+  between them from a no-JavaScript branch switcher. Branch creation is the
+  enabling piece: no path previously put a second `refs/heads/*` on a project
+  repo, so there was nothing to list. A new branch can only start from a commit
+  the repository already holds, so creating one can never introduce content.
+  Backups now record branch refs and restore rebuilds them.
+- **The diff's base commit reaches CI (#274).** Evaluation webhooks now carry
+  `baseSha`, the commit the diff was actually taken against, so a receiver can
+  apply the hunks to the right revision instead of guessing at whatever the
+  default branch happened to be when the request landed.
 - **Per-user telemetry opt-out.** Settings → Privacy now has a switch to stop sending product
   analytics for your account, alongside a plain-language disclosure of exactly what is sent.
   An agent inherits its owner's choice. Telemetry remains on by default; the existing
@@ -33,6 +45,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   onto a named token first — disabling cannot be undone.
 
 ### Fixed
+- Approvals are dismissed when the evaluated **base** moves, not only when the
+  tip does. A change re-evaluated against a newer base kept approvals that were
+  granted against different code.
 - `STRATUM_TELEMETRY_DISABLED` had no effect on `deploy:production` or `deploy:staging`. It was
   declared only under top-level `[vars]`, which named wrangler environments replace rather than
   inherit, so self-hosters who set it were still sending telemetry. It is now declared per
@@ -49,6 +64,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   than under a name.
 
 ### Security
+- **Webhooks are scoped on project id alone (#235).** `webhookBelongsToProject`
+  and `listWebhooks` still matched `project_id = ? OR (project_id IS NULL AND
+  project = ?)`, so a pre-migration-025 row could be read, modified and
+  delivered to by a same-named project in another namespace. `listWebhooks` now
+  takes the project id rather than its name, so a name-scoped lookup is not
+  expressible at the call site, and `createWebhook` requires `projectId` so no
+  new unstamped row can undo the backfill.
 - Sandbox evaluation of a pinned commit no longer clones a workspace's entire reachable
   history into memory: `readRepoFiles` now clones shallow and grows the fetch window only
   as far as needed to reach the pinned commit, capped at 500 commits, instead of an
--- a/docs/api/endpoints/projects.md
+++ b/docs/api/endpoints/projects.md
@@ -26,3 +26,54 @@ body must confirm the exact path:
 Returns `202 Accepted` with `{ "status": "deleting", "jobId": "del_…" }` — the
 cascade runs asynchronously and is idempotent/resumable. A mismatched `confirm`
 returns `400`; a non-owner returns `404`.
+
+## List Branches
+
+`GET /api/projects/{namespace}/{slug}/branches`
+
+Returns `{ defaultBranch, branches: [{ name, oid }], truncated, totalBranchCount }`.
+Read from the remote's ref advertisement, so the cost does not grow with the
+branch count. Capped at 200 — `truncated` says so explicitly, and the default
+branch is never the entry dropped.
+
+## Create Branch
+
+`POST /api/projects/{namespace}/{slug}/branches`
+
+```json
+{ "name": "release/2.x", "startPoint": "trunk" }
+```
+
+`startPoint` may be a branch name or a full 40-character commit sha, and
+defaults to the default branch's tip. Short shas and **tag names** are refused:
+a tag can point at a non-commit, its peeled commit is not advertised on the push
+path, and a tag may sit outside both the default branch's history and the tag
+set a backup captures — so a branch created from one could fail to push or
+vanish on restore. The new ref can only point at an object the repository
+already holds, so branch creation cannot introduce content that has not passed
+the change gate.
+
+**Writers only** (a non-writer gets `404`). `400` for an invalid name or an
+unresolvable start point. Three distinct `409`s: `BRANCH_EXISTS` if the name is
+taken, `BRANCH_NAME_CONFLICT` if it collides with an existing ref path (`release`
+when `release/2.x` exists — git stores refs as paths, so one cannot be both a
+file and a directory), and `NO_DEFAULT_BRANCH` if the configured default branch
+is not advertised by the remote and there is therefore nothing to default to.
+
+## Delete Branch
+
+`DELETE /api/projects/{namespace}/{slug}/branches/{name}`
+
+Hierarchical names are passed as real path segments
+(`.../branches/release/2.x`); a percent-encoded slash (`release%2F2.x`) names
+the same branch, since each path segment is decoded individually.
+**Writers only.** The default branch cannot be deleted —
+`409 DEFAULT_BRANCH_PROTECTED`.
+
+## Browsing a branch
+
+`GET .../files`, `.../content`, and `.../log` accept `?ref=<branch>`, defaulting
+to the project's default branch. Branch names only: an unknown ref is a `404`
+(never a silent fall back to the default), and a name that is both a branch and
+a tag is a `409 AMBIGUOUS_REF` rather than a guess at which was meant. The
+response echoes the `ref` actually read.
--- a/docs/api/openapi.yml
+++ b/docs/api/openapi.yml
@@ -493,6 +493,8 @@ paths:
       tags: [projects]
       summary: List files in the project repository
       security: [{ bearerAuth: [] }, {}]
+      parameters:
+        - $ref: "#/components/parameters/Ref"
       responses:
         "200":
           description: File listing
@@ -504,6 +506,9 @@ paths:
                   namespace: { type: string }
                   slug: { type: string }
                   path: { type: string }
+                  ref:
+                    type: string
+                    description: The branch actually read
                   files:
                     type: array
                     items: { type: string }
@@ -527,6 +532,7 @@ paths:
           required: true
           schema: { type: string }
           description: File path within the repository
+        - $ref: "#/components/parameters/Ref"
       responses:
         "200":
           description: File content, or a marker for binary/oversized files
@@ -539,6 +545,9 @@ paths:
                   namespace: { type: string }
                   slug: { type: string }
                   path: { type: string }
+                  ref:
+                    type: string
+                    description: The branch actually read
                   kind:
                     type: string
                     description: '"content" when `value` carries the text; other kinds (e.g. binary/too-large) omit `value`.'
@@ -566,6 +575,7 @@ paths:
           in: query
           schema: { type: integer, default: 20 }
           description: Number of commits to return
+        - $ref: "#/components/parameters/Ref"
       responses:
         "200":
           description: Commit log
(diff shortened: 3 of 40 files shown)


jlamoreaux:
All 12 review findings addressed in `d7db435` (the lint one landed earlier in `f50a59a`). Two notes where I did something other than the literal suggestion, and one thing I can't fix from the diff.

**Re-evaluation base (Major) — confirmed, and it was worse than described.** `UpdateChangeStatusOpts` had no `baseSha` field at all, so the re-evaluate route could not have re-pinned it even if it tried. Added the field, wired it into the shared status-statement builder (which both the plain update and the atomic dismiss+update batch go through), and pinned it to the `baseOid` the re-evaluation actually ran against. Worth flagging that this is **pre-existing** — `change.baseSha` was never refreshed on re-evaluation before this PR either — but it is squarely the class of bug #274 is about, and this PR is what put the correct value in scope at that point, so it belongs here.

**Repo-view cost (Major) — took the first option, plus lazy minting.** The token mint was the larger half of the regression and the suggestion only covered the switcher, so: the read token is now minted lazily and at most once, only when a ref was actually requested or the snapshot could not serve the page. On a snapshot hit the page makes **zero** git calls again, and the switcher offers just the branch being shown. There's a test asserting `freshRepoToken` and `listRepoBranches` are never called on that path, so the regression can't come back silently.

I did not take the second option (recording branch names in the KV snapshot). It would put a second, staler copy of the branch list behind a key with no branch component and no invalidation on branch create/delete — the same shape as the `?ref=` snapshot bug this PR exists to avoid.

**`branches: []` on a failed advertisement — agreed, and it needed a test to pin the distinction.** Passing `undefined` fixes the manifest, but nothing then distinguished "advertisement failed" from "genuinely zero branches" in the suite, so both cases are now asserted separately.

**Not a code fix: the linked-issues check on #235.** It asks for evidence that the backfill has run and left zero NULL `project_id` rows. That's an operational precondition for *deploying* this commit rather than something the diff can carry — `POST /api/admin/backfill/webhooks/apply` reports `remainingNullRows`, and it should read zero before this ships. Flagging it for @jlamoreaux rather than marking it done.

Gates re-run in CI's order: lint, typecheck, full suite (2350 passed / 27 env-gated skips), and `test:coverage` thresholds.

---

jlamoreaux:
@coderabbit review

jlamoreaux:
Second round addressed in `3af33eb`. Five of six taken; one declined with reasoning.

**The approvals/base finding is correct, and it is a bug this PR opened.** Worth spelling out because the review's framing understates it. Dismissal keyed only on `evaluatedSha` was *sufficient* while `baseSha` was frozen at creation: a change whose base advanced stayed permanently `STALE_BASE` and could not merge on an old approval at all. Re-pinning the base (#274, this PR) makes re-evaluation the documented remedy for staleness — and with it, an unchanged workspace tip against a moved base clears the merge gate while carrying approvals given for `diff(oldBase..tip)`, which is not the diff that would merge. So the same commit that fixed the un-mergeable-forever problem opened the laundering path. Dismissal now triggers on either endpoint moving, and the audit detail records which one, since a reader otherwise cannot distinguish a re-push from a base advance.

Two tests, not one: base-moved-tip-static dismisses, **and** the paired negative where neither moved keeps approvals. Without the second, the first would pass equally against an implementation that dismissed on every re-evaluation.

**Snapshot `branches` contract** — agreed, and this was my own inconsistency: the previous round changed the failed-advertisement path to pass `undefined` while leaving a docstring that said absence meant "no branches". Absence now means "not captured"; `[]` means the remote confirmed none. Restore deliberately collapses both to "tip only" and now says so — the distinction exists for a reader reporting on backup completeness, not for restore.

**The test-harness `any`** — taken. Making the capture arrays real types also turned indexed access into a genuine check, so reads go through a helper that names a missing capture rather than failing several lines later on a property of `undefined`.

**Docs** — both taken. `BRANCH_NAME_CONFLICT` and `NO_DEFAULT_BRANCH` were undocumented 409s, and the webhook comment conflated "available" with "serialized into a response".

**Declined: widening the start-point clone.** The mechanism is real — `cloneRepo` is `singleBranch`, so `fullHistory` covers the default branch only, and a raw sha reachable only from another branch is reported `bad-start-point`. But it fails *closed*, and no path can currently produce such a commit: every branch this API creates points at the default tip, another branch's tip, or a sha already in default history, so branch tips stay reachable from the default branch by induction, and import and restore each write a single branch's history. Widening the fetch to every advertised branch multiplies a clone that lands wholly in worker memory, against a ~1000-subrequest budget, to validate one sha. Recorded in the docstring along with the two changes that would make it reachable (a force-push gate, or a default-branch switch).

Gates re-run in CI's order: lint, typecheck, full suite (2352 passed / 27 env-gated skips), and `test:coverage` thresholds.

@jlamoreaux — one thing still needs you rather than the diff, unchanged from my earlier note: the #235 backfill is an operational precondition for deploying this commit. `POST /api/admin/backfill/webhooks/apply` reports `remainingNullRows`, and it should read zero before this ships.

---

jlamoreaux:
**`Deploy to Staging` is failing, and it is not this PR's diff — it is my own PR #322. Flagging rather than fixing here, because the fix is not portable into this branch.**

```
✘ [ERROR] A request to the Cloudflare API (/accounts/***/workers/scripts/stratum-staging) failed.
```

#319 does not touch `wrangler.toml`, `src/index.ts`, or any Durable Object.

### What actually happened

`pr-checks.yml`'s `deploy-staging` runs `npx wrangler deploy --env=staging` on **every** pull request, and every PR targets the **same** worker, `stratum-staging`. Its concurrency group is `deploy-staging-${{ github.head_ref }}` — per branch, not per environment — so PRs deploy over each other by design.

| PR | `Deploy to Staging` | migration tags in that branch's `wrangler.toml` |
|---|---|---|
| #321 | ✅ 16:39:20–16:39:49 | v1, v2 |
| #322 | ✅ 16:39:53–16:40:24 | v1, v2, **v3** |
| #319 | ❌ 19:48 | v1, v2 |

#322 adds `[[migrations]] tag = "v3"` for a new `MagicLinkRateLimiter` Durable Object. Its 16:40 deploy moved the live `stratum-staging` worker to migration tag **v3**. Every other branch's config tops out at **v2**, so wrangler cannot find the worker's current tag in its local migration list, cannot compute a delta, and falls back to applying from the beginning — where `new_classes = ["MergeQueue"]` collides with the DOs already living there.

This PR's staging deploy was green at 14:43 with the identical `wrangler.toml`. The only variable that changed between then and 19:48 is #322's deploy.

### Consequences worth stating

- **#321 is not actually safe either.** It is green only because it deployed four minutes *before* v3 landed. Its next staging deploy fails the same way.
- A re-run will not help: this is deterministic remote worker state, not a flake. I have not spent one.
- This recurs for **every** future PR that adds a Durable Object migration, not just this one.

### What clears it

Either merge #322 (after which any branch rebased on `main` carries v3 and the tags line up again), or rebase #319 and #321 onto a `main` that contains it. I have deliberately not ported #322's migration into this branch: declaring `MagicLinkRateLimiter` here without its source file would just trade this error for a missing-export one.

### The underlying issue, for a separate change

Running a shared-environment `wrangler deploy --env=staging` from every PR makes any DO migration a cross-PR blocker. `pr-preview.yml` already gives each PR its own isolated worker (`stratum-pr-319`) and database, and that job is green here — so the shared staging deploy is largely redundant on pull requests. Restricting `deploy-staging` in `pr-checks.yml` to `main` (as `ci.yml`'s own `deploy-staging` already is) would remove the interference class entirely. That is a deploy-infrastructure change well outside this PR's scope, so I am proposing it rather than pushing it — happy to open it as its own PR if you want it.

Everything else on this head is green: lint, typecheck, unit tests, integration tests, security scan, packages, docs, and the per-PR preview environment.

---

PR #321, opened 2026-08-28
Scoped, expiring API tokens (#254)

jlamoreaux:
## Summary

Replaces the single `users.token_hash` credential with named, scoped, expiring tokens. A leak was previously full account power forever, CI could not be given anything weaker than the whole account, and rotating for one consumer broke every other.

### Enforcement keys on resolved scope, not on paths

A read-only token is refused in exactly two places, both of which every write already passes through:

- **`authMiddleware`**, before routing, so no write route has to remember the check and one added later inherits it. It is an **allow-list** — only `GET` and `HEAD` are reads — because a deny-list of the four common write verbs fails *open* on the fifth. A test asserts an unrecognised verb is treated as a write.
- **`authorizeProject` / `authorizeWorkspace`** in git-http, which already compute `"read" | "write"`.

That second point is the correction that matters. My first design said "allow `info/refs`, refuse `git-receive-pack`" — but `GET /info/refs?service=git-receive-pack` authorizes with `canWriteProject` and proxies with a **write-scoped Artifacts token**. A path rule would have handed a read-only token a write authorization and a minted write credential. Keying on the resolved scope covers all **four** write entry points — project and workspace `git-receive-pack`, and both advertisements. Verified by deleting the two-line guard: exactly those four fail, nothing else.

A read-only token still clones, which needs stating because `git-upload-pack` is a POST — a method rule alone would have blocked precisely the CI user this feature exists for.

### A scoped token cannot mint a credential that outlives it

Token management is session-only, so a `read_write` token cannot mint siblings or revoke them. That guarantee had a hole, found by the AI reviewer in the round below and fixed in `bcd8ce3`: **`POST /settings/rotate-token` and `POST /api/users/me/rotate-token` still minted the *legacy* key**, which never expires and cannot be revoked one at a time. Revoking a compromised scoped token therefore contained nothing if it had rotated the legacy key on the way out — the exact circularity `requireSettingsSession` exists to prevent, reached through two routes that helper does not cover.

My original defence of those routes was backward compatibility. It only ever covered credentials that **predate** this PR: sessions and the legacy key keep rotating, so no existing automation breaks, and scoped tokens are new here so nothing can yet depend on them reaching these routes.

`tokenScope` could not express the distinction — the legacy credential also resolves to `read_write` — so `authMiddleware` now publishes `apiTokenId`, set for scoped tokens **only**, and `isScopedTokenCaller` keys on it. Two tests pin the discriminator itself: a scoped token sets `apiTokenId`, the legacy key leaves it unset. Without those, the middleware could stop setting it and every guard would silently pass everyone with nothing red.

### Deployable without an outage, and then closable

Authentication tries scoped tokens, then the legacy credential. A scoped token that resolves is answered there — revoked, expired, or a soft-deleting owner all reject rather than falling through, so the fallback cannot launder a rejected credential. A miss, or a storage failure reading the table, tries `users.token_hash`, which the caller still has to satisfy on its own hash. `POST /me/legacy-token/disable` then lets a user turn the old credential off, so the guarantee is available to existing accounts and not only new ones.

### Things that would have been silent bugs

- **Account erasure.** `api_tokens.user_id` references `users(id)`, so without a cascade delete the final `DELETE FROM users` throws, the job records a failure and retains the row, and erasure never completes for anyone who has minted a token. The existing account-cascade suite uses a statement-recording stub that evaluates no constraints, so it would have stayed green. The new test enables `PRAGMA foreign_keys = ON` explicitly and was verified to fail without the fix.
- **Expiry comparison.** Timestamps are ISO 8601 and compared **in JavaScript, never in SQL**. This codebase stores both `datetime('now')` (`"… 12:00:00"`) and ISO (`"…T12:00:00.000Z"`), and `' '` sorts below `'T'` — so `WHERE expires_at > datetime('now')` against an ISO value reads every expired token as *unexpired*. Sessions already dodge this the same way. Now factored into a shared `isExpired`, because the rule had to hold in a third place (below).
- **The active-token cap counted expired rows.** `revoked_at IS NULL` alone meant a user whose 20 tokens had all lapsed was locked out until they revoked twenty dead rows by hand — the precise lockout the cap's own docstring says revoked rows are excluded to avoid. The settings counter had the same bug. Fixed in `bcd8ce3`.
- **`c.executionCtx` throws** rather than returning undefined when none was supplied, which is every `app.fetch(request, env)` test here. `last_used_at` goes through `getWaitUntil`, debounced to at most hourly so it is affordable enough to actually base a revocation decision on.
- **`GET /settings` now requires a session.** An API token could previously render the page and read token metadata from the HTML while the JSON routes refused tokens — the page was the weaker of two doors to the same room.

### Decisions taken (both confirmed with the maintainer)

- **Token management is session-only.** GitHub forbids PATs from managing PATs for the same reason.
- **An explicit legacy-token disable**, rather than leaving the old credential live until the column is dropped.

Creation defaults to the **weaker** scope: a caller who does not say what they need should not be handed the ability to write.

## Related issues

Part of #254 — this is the **read/write** half. The `project/org` dimension the issue also asks for is deferred to #323, so this deliberately carries no closing keyword: #254 stays open until #323 lands.

## Type of change

- [x] New feature
- [x] Bug fix
- [x] Documentation

## How was this verified?

- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test` — 2448 passed, 27 env-gated skips, 0 failures
- [x] `npm run test:coverage` — thresholds pass

All four confirmed by **exit code**, not by reading tailed output — a lint error once shipped from this session because `| tail` hid it.

The storage and cascade tests run against **real SQLite with the production migrations applied**, not a statement-matching stub, because the behaviour that matters there is decided by the schema: the `CHECK` on `scope`, and the foreign key the cascade has to satisfy. Every guard in this PR was verified by reverting it and watching exactly the right tests fail — including both fixes in `bcd8ce3`.

## Review rounds

**CodeRabbit** posted 7 findings; all 7 fixed in `b4c24de`. One was a real bug: `POST /api/users/me/tokens` reported a storage-side **400 as a 409**, telling a caller to free a token slot to fix a rejected name. Also fixed: `SETTINGS_NOTICES[query]` resolved inherited names, so `?notice=constructor` rendered an empty notice card; the token id lived in the free-form `detail` blob for the JSON routes but in the indexed `subject` column for the settings routes, so an audit query keyed on `subject` returned half the token lifecycle; a redundant `idx_api_tokens_hash`; a wholesale users-module mock that dropped exports `src/index.ts` imports; and two docs/import nits.

**PR-Agent** (`/review`) found the two bugs fixed in `bcd8ce3` and described above. Both were real, both are now covered by tests verified to fail against the pre-fix source. Its containment finding is the more valuable of the two — it caught that my own stated invariant did not hold on two routes I had argued my way past.

Worth noting for the next reviewer: the settings-page test mocked `../src/storage/api-tokens` by **listing** its exports, so adding `isExpired` made it arrive as `undefined` and 500 the render. That mock now spreads the real module and overrides only what touches D1 — the same defect class CodeRabbit flagged earlier in this PR.

## Explicit non-goals, and known gaps

**Per-project and per-org scopes are not here — tracked as #323.** `canReadProject`/`canWriteProject`/`isProjectAdmin` back ~70 call sites across nine route files, and a scope honoured in only some of them reads as a guarantee it is not — a token the UI labels "scoped to project X" that actually works everywhere is *more* dangerous than an honestly unscoped one, because it invites the operator to hand it to a third party. #323 records the design point that makes it tractable: make the binding a **required** parameter of the `authz.ts` helpers, so every un-updated call site becomes a compile error rather than a silent bypass. `scope` is a TEXT column with a `CHECK` so the vocabulary can grow without a migration rewrite.

Consequences worth stating rather than discovering:

- **`POST /settings/agents` and `POST /api/agents` are still reachable by a `read_write` token.** The reviewer flagged these alongside rotate-token, and they are deliberately left. The asymmetry that makes rotate-token urgent does not apply: an agent token **is** individually revocable — deleting the agent kills it — so "revoke the lost laptop" still contains the breach. The legacy key is neither revocable individually nor expiring, which is why it alone is gated here. Narrowing agent creation belongs with the deferred scoping work.
- `POST /api/users/me/delete` is likewise still token-reachable.
- **"Read-only" bounds damage, not exposure.** Such a token still reads every private repo its owner can read, and for an admin-matched email the audit log and metrics too.
- Nothing rate-limits a failed token: `authMiddleware` runs before `rateLimitMiddleware`, and `git-upload-pack` is exempt entirely. Unchanged by this work — 128 bits of entropy is the defence, and this does not pretend to add one.
- `src/evaluation/secret-scanner.ts` matches `stratum_user_[a-f0-9]{32}`. The 16-byte `generateApiKey` format is load-bearing for leak detection; a test pins it.

## Checklist

- [x] Docs updated where the change makes them inaccurate — `openapi.yml` and `authentication.md`, including the rotate-token contract change in `ee9500a`
- [x] The web UI change stays server-rendered with no client-side JavaScript


jlamoreaux:
**`Deploy to Staging` is now failing here too. Not this PR's diff — it is my own PR #322. Flagging rather than fixing, because the fix is not portable into this branch.**

This is the failure I predicted on #319 (comment 5457080746) would reach this PR. Its earlier green was not a sign of health: it deployed at 16:39:20, four minutes *before* #322's deploy moved the shared staging worker to Durable Object migration tag **v3** at 16:40:24. The merge commit `1a0b455` is simply the first staging deploy this branch has attempted since.

**Cause.** `pr-checks.yml`'s `deploy-staging` runs `wrangler deploy --env=staging` on **every** pull request, all against the **same** worker (`stratum-staging`), with a per-*branch* concurrency group. #322 adds `[[migrations]] tag = "v3"` for a new Durable Object. This branch's `wrangler.toml` tops out at **v2**, so wrangler cannot find the live worker's tag in its local migration list, cannot compute a delta, and falls back to applying from the beginning — where `new_classes = ["MergeQueue"]` collides with the Durable Objects already there:

```
```

**What I verified versus what I am inferring.** I confirmed that exact error twice on #319, once on `3af33eb` and again on a deliberate re-run at 20:46 — deterministic, not a flake. For *this* job the log tail available to me is truncated past the wrangler step, so I have not read the error line here. The cause is structurally identical (same job, same shared worker, same v2-capped config), but I would rather say that plainly than claim a confirmation I do not have.

I have **not** spent a re-run here. Determinism is already established on the identical cause, so a re-run would burn a deploy to learn nothing.

**What clears it:** merging #322, or rebasing this branch onto a `main` that contains it. I have deliberately not ported #322's migration here — declaring `MagicLinkRateLimiter` without its source file only trades this error for a missing-export one.

**The underlying issue, for a separate change:** running a shared-environment staging deploy from every PR makes any Durable Object migration a cross-PR blocker, and this will recur for every future one. `pr-preview.yml` already gives each PR an isolated worker and database (`stratum-pr-321`, green here), and `ci.yml`'s own `deploy-staging` is already restricted to `main`. Restricting `pr-checks.yml`'s copy the same way removes the interference class entirely. That is deploy infrastructure well outside this PR, so I am proposing it rather than pushing it.

Everything else on `1a0b455` is green: lint, typecheck, unit tests, integration tests, security scan, packages, docs, and the per-PR preview environment.

---

jlamoreaux:
Both PR-Agent findings were real. Fixed in `bcd8ce3`, docs in `ee9500a`.

**Containment gap — fixed.** The finding is correct and it caught my own stated invariant failing. `requireSettingsSession`'s docstring says a `read_write` token must not be able to issue itself a replacement, and then `POST /settings/rotate-token` and `POST /api/users/me/rotate-token` let it mint the *legacy* key — which never expires and cannot be revoked individually. Revoking a compromised scoped token contained nothing if it had rotated that key first.

My defence of those routes was backward compatibility, and the reviewer is right that it does not stretch to cover scoped tokens: they are new in this PR, so nothing can depend on them reaching these routes. Sessions and the legacy credential still rotate, so no existing caller breaks.

`tokenScope` could not carry the distinction — the legacy credential also resolves to `read_write` — so `authMiddleware` now publishes `apiTokenId`, set for scoped tokens only, and `isScopedTokenCaller` keys on it. Two tests pin the discriminator itself (scoped sets it, legacy leaves it unset), because otherwise the middleware could stop setting it and every guard would pass everyone with nothing red.

**Cap counting expired tokens — fixed.** Exactly the lockout the cap's own docstring says revoked rows are excluded to avoid. Counted in JS, not SQL, for the mixed-timestamp reason the finding correctly flagged; factored into a shared `isExpired` now used by `resolveApiToken`, the cap, and the settings counter.

**`POST /settings/agents` — deliberately not changed.** The pattern looks identical but the asymmetry that makes rotate-token urgent is absent: an agent token *is* individually revocable, since deleting the agent kills it, so "revoke the lost laptop" still contains the breach. The legacy key is neither revocable individually nor expiring, which is what makes it the one credential worth gating here. Narrowing agent creation is already recorded as deferred work under #323 rather than widened into this PR.

Both fixes were verified by reverting them and confirming exactly the new tests fail. Gates confirmed by exit code: biome 0, tsc 0, `vitest --coverage` 0 — 2448 passed, 27 env-gated skips.

One note for whoever touches these tests next: the settings-page suite mocked `../src/storage/api-tokens` by listing its exports, so adding `isExpired` made it arrive as `undefined` and 500 the render on first run. That mock now spreads the real module and overrides only what touches D1 — the same defect class flagged earlier in this PR.

---

PR #322, opened 2026-08-28
Make magic-link rate-limit admission atomic with a Durable Object (#283)

### LINKED ISSUE #283: Make magic-link rate-limit admission atomic with a Durable Object
## Summary

Replace the non-atomic Workers KV read-modify-write flow used by magic-link rate limiting with a serialized Durable Object counter or an equivalent serialized reservation mechanism.

The current implementation is an intentional partial mitigation. Concurrent requests can read the same counter values and overwrite each other's increments. This can allow requests to exceed the per-email or per-IP limits.

## Required changes

- Migrate both magic-link counters together:
  - Per-email limit: 5 sends per hour.
  - Per-IP limit: 20 sends per hour.
- Keep `checkMagicLinkRateLimits()` in `src/routes/email-auth.tsx` as the shared call-site boundary, or provide an equivalent single admission API for all magic-link send flows.
- Use a Durable Object or another serialized mechanism to reserve admission before a send occurs.
- Do not migrate only the per-IP counter. Both counters must have the same consistency guarantee.
- Decide and document the storage-error policy in the same change:
  - The current policy fails open to avoid blocking login during a transient storage outage.
  - A fail-open policy permits bypass during storage errors.
  - A fail-closed policy can cause an authentication outage.
  - The implementation must make this trade-off explicit rather than changing it implicitly.

## Tests

Add concurrency tests that exercise the admission/read-to-write race window. Sequential send tests are insufficient because they already pass with Workers KV counters.

The tests must verify that concurrent requests from one source IP cannot admit more than 20 sends in one hour. Add equivalent coverage for the per-email cap if the new admission API supports testing it directly.

## Affected area

- `src/routes/email-auth.tsx`
- Magic-link rate-limit storage and test infrastructure

## Rationale

The current Workers KV implementation is a scoped partial mitigation for the magic-link mail-bombing and email-cost amplification vector. It bounds normal traffic but cannot atomically enforce limits under concurrency.

## Backlinks

- Originating pull request: https://github.com/stratum-eng/stratum/pull/231
- Originating review discussion: https://github.com/stratum-eng/stratum/pull/231#discussion_r3863067659
- Requested by: @jlamoreaux

jlamoreaux:
## Summary

The per-email (5/hour) and per-IP (20/hour) magic-link caps were Workers KV counters, read and written from the Worker:

```ts
emailCount = Number.parseInt((await c.env.STATE.get(emailKey)) ?? "0");
...
await c.env.STATE.put(emailKey, String(emailCount + 1), ...);
```

Concurrent sends read the same value and each wrote back `value + 1`, so N simultaneous requests advanced the counter by **one**. The caps bounded sequential traffic only — which is exactly what the originating review said they were: a scoped partial mitigation for mail-bombing and email-cost amplification.

Both counters move together to a `MagicLinkRateLimiter` Durable Object, one instance per subject (`email:<sha256>` or `ip:<addr>`). The runtime's input gate holds other events for an object while one of its storage operations is in flight, so the read and the write cannot interleave.

### The admission API now decides and commits in one step

`checkMagicLinkRateLimits()` stays the single call-site boundary the issue asked for, but the reservation *is* the decision — there is no longer a window between "am I under the cap" and "count me". This costs nothing at the call sites: all three already invoked `commit()` as the first statement after an unblocked check, with nothing in between that could change the outcome. `commit()` is gone.

### Why two objects, and what the refund is for

The two caps partition differently — per address and per source — so no single instance can hold both without becoming a global serialization point for every magic-link send on the platform. They are therefore reserved in sequence (email, then IP), and the email reservation is **refunded** when the IP cap then refuses.

Between the reservation and the refund, a concurrent request can observe the un-refunded count and be turned away. That direction — over-refusing, briefly — is the safe one, and **neither cap is ever exceeded**, which is the invariant that matters. A refund that itself fails leaks one count against the window, which can only refuse sends, never admit extra ones.

### Storage-error policy: fail open, stated rather than changed

The issue asked for this to be decided explicitly. It stays **fail open**: a reservation that cannot be *attempted* — the binding is absent, or the object throws — admits the request and logs it. Failing closed would turn a transient storage fault into a total login outage, which is a worse failure than the bounded over-sending a fault window allows; both caps re-apply on the next request once storage recovers.

Two things that are deliberately *not* covered by that:

- A reservation that **succeeds and reports the cap** still refuses. That is an answer, not a fault. The KV version admitted there too, but only as an accident: a throw on the second read reset the first read's count as well. The old "fails open when the per-IP read throws after the email bucket is full" test encoded that accident, and is replaced by one asserting the actual contract against both counters.
- A **malformed limit or window refuses** rather than throwing. A throw would reach the caller's `catch`, which fails open — turning a typo in a constant into a silently unlimited endpoint.

An absent `MAGIC_LINK_LIMITER` binding (only reachable on a self-hosted deploy whose `wrangler.toml` predates it) logs at ERROR, because it means the endpoint is running with no cap at all.

### Storage lifecycle

Each object arms an alarm past the end of its live window and `deleteAll()`s on it, which is what `expirationTtl` did for the KV keys — a subject seen once does not leave durable state behind. The hour is no longer part of the object name (it lives inside the object, which is what makes self-erasure possible), so in-flight hourly counters reset once on deploy.

## Related issues

Closes #283

## Type of change

- [x] Bug fix
- [x] Refactor / cleanup
- [x] Documentation

## How was this verified?

- [x] `npm run typecheck`
- [x] `npm test` — 2251 passed, 27 env-gated skips, 0 failures
- [x] `npm run lint`

18 new tests. The concurrency tests run against a fake namespace that models workerd's input gate with a per-instance promise chain; storage accessors are async, so an *ungated* instance genuinely interleaves.

**One test asserts that the same 50 concurrent reservations OVER-admit against an ungated instance.** Without it the cap assertion could pass for the wrong reason — a harness that accidentally stopped modelling concurrency would make every concurrency test in the file vacuous, and nothing would say so. The route-level tests cover the issue's stated requirement directly: 25 concurrent sends from one IP admit exactly 20, 12 concurrent sends for one address admit exactly 5, and 8 addresses × 6 sends from one IP admit exactly 20 (the tighter cap winning). Also covered: an email-capped request does not spend IP budget, an IP-capped request gets its email reservation back, exactly one reservation RPC is issued per counter (no Worker-side read-then-write), and both fail-open paths.

## Checklist

- [x] Docs updated where the change makes them inaccurate — the architecture doc's KV structure no longer lists the removed key, and a new section records the design and the fail-open policy

## Notes

**Not addressed here, on purpose:** the windows stay *fixed* rather than sliding. A fixed window permits a 2× burst across its boundary, but that is a property of the limit's shape, not of its atomicity, and changing it would change user-visible behaviour beyond what this issue asks for.

**Erasure:** a bucket is keyed by an unsalted SHA-256 of the address (as the object *name*; the stored value is only `{window, count}`) and erases itself within the hour. That matches the 1-hour KV TTL it replaces, so account deletion needs no new cascade — but it is worth knowing the property is the alarm, not a `purge()` RPC.

https://claude.ai/code/session_01P11maBt5wGR783noPFLpqs

---
_Generated by [Claude Code](https://claude.ai/code/session_01P11maBt5wGR783noPFLpqs)_

## Summary by CodeRabbit

- **New Features**
  - Added more reliable magic-link request limits based on both email address and IP address.
  - Applied limits consistently across signup, login, and legacy magic-link sending flows.
  - Requests can continue during temporary rate-limiter storage failures.

- **Bug Fixes**
  - Improved concurrent request handling to prevent accidental over-admission.
  - Added refunds for email reservations when IP limits block a request.
  - Rate-limit windows now reset and clean up automatically.

### DIFF
--- a/docs/developer/architecture.md
+++ b/docs/developer/architecture.md
@@ -177,7 +177,6 @@ CREATE TABLE github_pr_mappings (
 
 ```text
 session:{sessionId} → { userId, expiresAt }
-rate_limit:magic_link:{emailHash} → { count, resetAt }
 sync_status:{namespace}:{slug} → { lastCheckedAt, hasUpdates, commitsBehind }
 eval_cache:{workspaceId}:{evaluatorId} → { result, cachedAt }
 ```
@@ -426,6 +425,35 @@ async function reportEvaluationToGitHub(env, change, project, evaluation): Promi
 _Evaluation performed by [Stratum](https://stratum.dev)_
 ```
 
+## Magic-Link Rate Limiter
+
+Durable Object holding one fixed-window send counter per subject — an
+address digest (`email:<sha256>`, 5 sends/hour) or a source IP
+(`ip:<addr>`, 20 sends/hour). A send is admitted only when it clears both.
+
+The counters lived in Workers KV until issue #283. That was a
+read-modify-write from the Worker, so concurrent sends read the same value
+and each wrote back `value + 1` — the caps bounded sequential traffic only.
+Inside a Durable Object the runtime's input gate holds other events while a
+storage operation is in flight, so the read and the write cannot interleave.
+
+Two subjects mean two objects, so the pair is reserved in sequence and the
+email reservation is refunded when the IP cap then refuses. A concurrent
+request can see the un-refunded count and be turned away; over-refusing
+briefly is the safe direction, and neither cap is ever exceeded. A single
+shared object would make the pair atomic but serialize every magic-link send
+on the platform behind one thread.
+
+**Storage errors fail open, deliberately.** A reservation that cannot be
+attempted — the binding is missing, or the object throws — admits the request
+and logs it. Failing closed would turn a transient storage fault into a total
+login outage, which is worse than the bounded over-sending a fault window
+allows. A reservation that *succeeds* and reports the cap still refuses: that
+is an answer, not a fault.
+
+Each object arms an alarm past the end of its window and erases itself on it,
+which is what `expirationTtl` did for the KV keys.
+
 ## Merge Queue
 
 Durable Object that serializes merge operations per repository.
@@ -501,7 +529,9 @@ Producer (API route) → Queue → Consumer Worker
 ```text
 POST /auth/email
   ↓
-Generate token → Store in KV (15 min TTL)
+Reserve against the per-email and per-IP caps (Durable Object)
+  ↓
+Generate token → Store in D1 (15 min TTL, single-use at verify)
   ↓
 Send email via Cloudflare Email
   ↓
--- a/src/index.ts
+++ b/src/index.ts
@@ -43,6 +43,7 @@ import type { Env, ImportJobMessage, MessageBatch, SyncJobMessage } from "./type
 import { CSS } from "./ui/styles";
 import { createLogger } from "./utils/logger";
 export { MergeQueue } from "./queue/merge-queue";
+export { MagicLinkRateLimiter } from "./queue/magic-link-limiter";
 export { RepoDO } from "./queue/repo-do";
 
 const app = new Hono<{ Bindings: Env }>();
--- a/src/queue/magic-link-limiter.ts
+++ b/src/queue/magic-link-limiter.ts
@@ -1,0 +1,118 @@
+import { DurableObject } from "cloudflare:workers";
+import type { Env } from "../types";
+
+/** The single storage key holding this bucket's window id and its count. */
+const BUCKET_KEY = "bucket";
+
+/**
+ * How long after a window closes the object waits before erasing itself. A
+ * counter is worthless once its window has rolled, but the grace period keeps a
+ * steadily-used bucket from thrashing delete/recreate against clock skew
+ * between the alarm and the next request.
+ */
+const CLEANUP_GRACE_MS = 60_000;
+
+interface Bucket {
+  /** `floor(epochSeconds / windowSeconds)` — the fixed window this count belongs to. */
+  window: number;
+  count: number;
+}
+
+export interface ReserveOutcome {
+  admitted: boolean;
+  /** The count after this call. Reported for logging; callers gate on `admitted`. */
+  count: number;
+}
+
+/**
+ * A serialized fixed-window counter, one instance per rate-limit subject (an
+ * email-address digest, or a source IP).
+ *
+ * This exists because the counter it replaces was a Workers KV
+ * read-modify-write: concurrent magic-link sends read the same value and each
+ * wrote back `value + 1`, so N simultaneous requests advanced the counter by
+ * one and the cap bounded sequential traffic only. A Durable Object closes that
+ * window — the runtime's input gate holds incoming events while a storage
+ * operation is in flight, so the read and the write below cannot interleave
+ * with another request against the same instance.
+ *
+ * Serialization is per instance, and an instance is per subject, so magic-link
+ * sends for different addresses (and from different IPs) still run in parallel.
+ *
+ * `limit` and `windowSeconds` are supplied by the caller rather than baked in
+ * here so the policy numbers stay next to the endpoint that enforces them.
+ * That is safe only because DO RPC is reachable from Worker code alone — no
+ * request can name its own limit. Do not expose this class over `fetch`.
+ */
+export class MagicLinkRateLimiter extends DurableObject<Env> {
+  /**
+   * Atomically admits or rejects one send against this subject's cap.
+   *
+   * @param limit - Maximum admissions per window
+   * @param windowSeconds - Fixed window length
+   * @param nowMs - Caller's clock, so a single request stamps both of its counters identically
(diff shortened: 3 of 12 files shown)


jlamoreaux:
**Heads-up on a side effect of this PR that is not visible from its own CI.**

This PR's `Deploy to Staging` ran at 16:39:53 and succeeded, moving the shared `stratum-staging` worker to migration tag **v3** (`MagicLinkRateLimiter`). Every PR in this repo deploys to that same worker — `pr-checks.yml`'s `deploy-staging` runs `wrangler deploy --env=staging` unconditionally on pull requests, with a per-*branch* concurrency group.

The consequence: any branch whose `wrangler.toml` tops out at v2 can no longer deploy to staging. Wrangler cannot find v3 in its local migration list, cannot compute a delta, and retries from the top, where `new_classes = ["MergeQueue"]` collides with the Durable Objects already there:

```
```

#319 hit exactly this at 19:48 (full analysis and timeline: #319 comment 5457080746). #321 will hit it on its next staging deploy — it is green only because it deployed four minutes before v3 landed.

Nothing is wrong with this PR's code, and I am not changing it: the migration is required for the feature, and it is correctly scoped (`new_sqlite_classes`, declared once, inherited by all three environments). Merging this first makes the problem disappear for every branch rebased on `main` afterwards.

The durable fix belongs elsewhere: `pr-checks.yml` deploying to a shared environment from every PR makes any Durable Object migration a cross-PR blocker. `pr-preview.yml` already provisions an isolated per-PR worker and database (`stratum-pr-322`, green here), so the shared staging deploy is largely redundant on pull requests — `ci.yml`'s own `deploy-staging` is already restricted to `main`. Proposing it rather than pushing it, since it is deploy infrastructure well outside this PR.

---

jlamoreaux:
@coderabbit review

jlamoreaux:
All 7 addressed in `f29fb02`. Two were real defects; one pre-merge warning is a misread I want to correct rather than act on.

### The two that mattered

Both were in the test harness, which is worse than their `🔵 Trivial` / nitpick labels suggest: a harness bug does not fail a test, it silently stops the tests from meaning anything.

**The stub Proxy answered `.then`.** My `get` trap returned a call wrapper for *every* property name, which made a stub a thenable. Awaiting one — or letting one settle inside a promise chain — would invoke `then` as a continuation rather than resolving to the stub. Symbol keys had the same problem (`Symbol.toPrimitive`, inspection hooks), and those are probes, not RPC. Both now return `undefined`.

**Fake storage handed out live object references.** Real Durable Object storage serializes, so a caller cannot reach back into stored state through a handle it kept. A `Map` of live objects can: a test could mutate a bucket in place and pass against behaviour production does not have. Both directions now `structuredClone`.

Each is pinned by a test, and I verified both fail against the pre-fix harness by reverting the two changes and re-running — they do.

### The typing ones compound

Declaring the binding as `DurableObjectNamespace<MagicLinkRateLimiter>` (your `src/types.ts` finding) made the other two fall out: both hand-written stub casts in `email-auth.tsx` became unnecessary, and the test stub type is now *derived* — `Pick<MagicLinkRateLimiter, "reserve" | "refund">` — instead of restated. A signature change in the Durable Object is now a compile error at every call site rather than something a cast absorbs. The `types.ts` import is type-only, so the cycle with `magic-link-limiter.ts` is erased.

One deviation worth flagging: you suggested putting the shared namespace construction in `fake-durable-object.ts`. I put it in a new `tests/helpers/magic-link-limiter.ts` instead, so the generic harness stays generic — it should not depend on one specific Durable Object class. That file also carries the single remaining cast, which the generic helper genuinely cannot avoid (it cannot know the RPC brand of whichever class it is handed). One cast in one place beats four.

The caps are now imported rather than copied, so a limit changed in one place cannot leave the tests asserting one nothing enforces. And `BUCKET_KEY`'s comment now says *why* one key holds both fields (a rollover overwrites, instead of accumulating a row per hour that something would have to sweep) rather than restating what the constant is.

### Correcting the Out of Scope warning

> The reported migration of generated magic-link token storage from KV to D1 is not required by the linked issue

No such migration is in this PR. `src/storage/` is untouched — `git diff da95544..HEAD -- src/storage/` is empty. Magic-link tokens already lived in D1 before this branch existed; `createMagicLink(c.env.DB, …)` has been the call for some time.

What the check is seeing is a **one-line documentation correction**. The architecture doc's auth-flow diagram claimed `Generate token → Store in KV (15 min TTL)`, which was simply false. I had to edit that same diagram anyway to insert the reservation step, and leaving a known-wrong line directly above a line I had just corrected seemed worse than fixing it. Happy to split it out if you would rather the diff be strictly minimal, but there is no code change to remove.

The Docstring Coverage warning I have left alone, per the standing preference.

Gates re-run in CI's order: lint, typecheck, full suite (2253 passed / 27 env-gated skips), and `test:coverage` thresholds.

---

jlamoreaux:
@coderabbit review

PR #324, opened 2026-08-29
Deploy shared staging from main only, and guard the migration chain

jlamoreaux:
## Summary

Every push to `main` failed from 16:38 UTC on 2026-08-28 until #322 landed ([run 33206494593](https://github.com/stratum-eng/stratum/actions/runs/33206494593)):

```
▲ [WARNING] The published script stratum-staging has a migration tag "v3",
  which was not found in your wrangler.toml file. [...] Applying all
  available migrations to the script...
✘ [ERROR] Cannot apply new-class migration to class 'MergeQueue' that is
  already depended on by existing Durable Objects [code: 10074]
```

The `MergeQueue` error is the symptom. `pr-checks.yml` ran `wrangler deploy --env=staging` on **every** PR, publishing the PR branch over the single shared `stratum-staging` Worker. Cloudflare records the last-applied Durable Object migration tag **on the script**, not in our config, so #322 — which adds a third migration for `MagicLinkRateLimiter` and was still open at the time — left staging at `v3` while `main`'s `wrangler.toml` stopped at `v2`. Wrangler could not find `v3` in the config, assumed it had been deleted, and replayed the chain from `v1`, which fails on the already-populated `MergeQueue` class.

A tag cannot be un-applied from a deployed script, so it did not clear on a re-run: `main` stayed red, and with it the production deploy that gates on staging. The commit in the failing run was unrelated — any push would have failed identically.

#322 has since merged, which reconciled the two. This PR stops it recurring, in two parts.

### 1. PRs no longer deploy to the shared staging Worker

The `deploy-staging` job is removed from `pr-checks.yml` rather than re-pointed, because per-PR environments already exist and are healthy: `pr-preview.yml` gives each PR an isolated `stratum-pr-<N>` Worker with its own D1 and KV, torn down when the PR closes. The shared environments are now deployed from `main` alone (`ci.yml`), which makes their migration tag a function of `main`'s `wrangler.toml` and nothing else.

### 2. The migration chain is append-only

Isolation alone does not close the hole. The same failure reaches **production** the moment a merged migration is reverted — an ordinary thing to do, and `ci.yml` gates production on staging, so one revert takes down both.

`tests/wrangler-migration-chain.test.ts` adds the missing invariant: a branch may append migrations, but may not remove, rename, reorder, or rewrite one that `main` already carries. It compares `wrangler.toml` against `origin/main` (hence `fetch-depth: 0` on the test job) and **fails rather than skips** in CI when it cannot reach `main` — a guard that quietly does not run is not a guard.

It rides the existing `Unit Tests` job, so it is already inside the `PR Requirements` aggregate check.

## Related issues

None filed — this came out of triaging the failing run directly.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [x] Other: CI/CD correctness

## How was this verified?

- [x] `npm run typecheck`
- [x] `npm test` — 2279 passed, 27 env-gated skips, 0 failures
- [x] `npm run lint`

The guard's parser and rule are unit-tested against fixtures (17 cases): comment and whitespace handling, a `#` inside a quoted value, per-environment `[[env.*.migrations]]` tables correctly ignored, and every violation shape — removal, rename, reorder, mid-chain insertion, redefinition, duplicate tag, missing tag.

Beyond the fixtures, the guard was run against a reproduction of this exact regression: deleting the `v2` block from `wrangler.toml` makes it fail with

> This branch removes migration(s) "v2" that are already on main. Those tags are recorded on the deployed staging and production scripts; dropping them makes wrangler replay the chain from the start and every deploy of those environments fails with error 10074.

Rebased onto `main` after #322 merged and re-verified: the chain is now `v1, v2, v3` on both sides and the guard passes.

## Checklist

- [x] Docs updated where the change makes them inaccurate — `docs/STAGING_SETUP.md` and `docs/DEVELOPER_WORKFLOW.md` both described PRs as deploying to staging; both now describe the per-PR preview environment, and each carries a section explaining why PRs stay off the shared Worker

## Notes

**Possible branch-protection follow-up:** if **"Deploy to Staging"** is a required status check on `main`, it needs removing from branch protection — the job no longer exists, so the check will never report. The `pr-status` job's `PR Requirements` aggregate never depended on it, which suggests it is not required, but I could not read protection settings to confirm.

**Two adjacent problems found while investigating, deliberately left alone:**

- `pr-checks.yml`'s `security-scan` job runs `git diff origin/main..HEAD` at the default shallow depth, where `origin/main` does not exist. With `continue-on-error: true`, the secret-in-diff check has likely never actually run.
- `wrangler.preview.toml` binds preview Workers to the **staging** Artifacts namespace, so PR previews share git object storage with staging — the same class of shared-state coupling this PR removes for the Worker script.

---
_Generated by [Claude Code](https://claude.ai/code/session_011JfBYHd2gURXXMNov8VT4N)_

## Summary by CodeRabbit

- **New Features**
  - Pull requests now use isolated preview environments, preventing changes from overwriting shared staging.
  - Staging deploys from the main branch, with production releases requiring successful staging validation and approval.

- **Documentation**
  - Updated workflow and deployment guidance with preview access, testing instructions, migration troubleshooting, and recovery procedures.

- **Tests**
  - Added safeguards to detect invalid or inconsistent migration sequences before deployment.
  - Improved CI validation of deployment history and migration changes.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -115,6 +115,10 @@ jobs:
           # migrations. The default shallow fetch has no origin/main, and the
           # guard fails rather than skips in CI, so fetch the full history.
           fetch-depth: 0
+          # The guard only reads local objects (git show, git rev-parse) and
+          # nothing else here talks to a remote, so the job needs no token in
+          # .git/config — and this job uploads an artifact.
+          persist-credentials: false
 
       - name: Setup Node
         uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -109,6 +109,12 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
+        with:
+          # tests/wrangler-migration-chain.test.ts reads wrangler.toml at
+          # origin/main to prove this PR only appends Durable Object
+          # migrations. The default shallow fetch has no origin/main, and the
+          # guard fails rather than skips in CI, so fetch the full history.
+          fetch-depth: 0
 
       - name: Setup Node
         uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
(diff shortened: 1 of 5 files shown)


jlamoreaux:
@coderabbit review

jlamoreaux:
@coderabbit review

jlamoreaux:
@coderabbit review

jlamoreaux:
@coderabbit review

PR #335, opened 2026-08-29
Send the diff's base commit to webhook evaluator receivers (#274)

### LINKED ISSUE #274: Webhook payload carries no base commit, so a receiver can evaluate the wrong tree

jlamoreaux:
Closes #274.

## The problem

`WebhookEvaluator` posted `{ diff, policy }` and nothing else. The diff is computed against the project's default-branch tip, but the identity of that tip never left the Worker — so the natural receiver (keep a mirror, check out the branch, `git apply`, run the suite) evaluates against whatever the branch points at when the request lands. With `merge.requiredEvaluators: ["webhook"]`, that is a merge gate that can silently judge the wrong tree, and neither end can tell.

## The fix

`getDiffBetweenRepos` already resolved the base tip so both sides of the diff read at a pinned commit — it just threw the value away. This returns it as `baseSha` and threads it to evaluators through a new optional `EvaluationContext`, which `WebhookEvaluator` serializes into the body.

```
POST { diff, baseSha, policy }   +   X-Stratum-Signature over the whole body
```

Wired at all three entry points: change creation (`change-flow.ts`, which also serves the gated `git push` front door), re-evaluation (`routes/changes.ts`), and manual conflict resolution (`routes/sync-management.ts`).

## One deliberate departure from the issue

The issue suggests forwarding `change-flow.ts`'s existing `baseSha`. **That is the wrong value.** It comes from `resolveProjectHead` — a KV-snapshot read taken *before* the diff clone — and the two can disagree. Forwarding it would assert a base the diff was demonstrably not computed against, replacing an honest absence with a confident lie.

This is guarded, not assumed: implementing the issue's literal suggestion fails the new regression test with `expected 'sha_base' to be 'diff_base_sha'`. The manual-resolution assertion was mutation-checked the same way (`expected undefined to be 'resolution_base_sha'`).

## Compatibility

`JSON.stringify` drops an `undefined` value, so an unpinned base produces the **byte-identical body — and therefore the identical HMAC** — that shipped before this field existed. The field is omitted, never `null`.

⚠️ **A receiver that rejects unknown fields** (zod `.strict()`, JSON Schema `additionalProperties: false`, serde `deny_unknown_fields`) will 4xx the new payload, which Stratum scores 0 — and under `merge.requiredEvaluators: ["webhook"]` that blocks every merge in the project. Both the changelog entry and the guide say to allow unknown fields before upgrading.

## Docs say what the field does *not* buy

Naming the base does not make the diff a faithful patch, and the guide now says so rather than implying reproducibility: added files are labelled `mode 100644` whatever their real mode (so executables and symlinks are misreported), renames appear as delete+add, binaries are decoded as text, and an unreadable blob renders as a spurious addition (fails `git apply` loudly) or a spurious deletion (**applies cleanly** — the dangerous direction). Advice is to pin the mirror to `baseSha` and report a failed apply as "cannot evaluate."

Also adds a one-line staleness banner to `docs/developer/architecture.md`, whose Evaluation Engine sketch documents an `Evaluator`/`EvalContext` interface that matches nothing in `src/` — and which a real `EvaluationContext` would otherwise collide with.

## Verification

| Gate | Result |
|---|---|
| `npm run typecheck` | clean |
| `npm test` | **2297 passing** (+8 new) |
| `npm run lint` | clean, 322 files |
| `npm run release:check` | passes |

⚠️ Two things reviewers should know:

- **The baseline is already red on `main`.** 43 failures in `tests/issues.test.ts` and `tests/issues-comment-body-limit.test.ts`, all `column index out of range` / `db.prepare is not a function` from the `node:sqlite` D1 shim on Node v22.14.0. Unrelated to this change; the failing set here is byte-identical to the baseline, with zero new failures. Worth its own issue.
- **Coverage thresholds could not be evaluated locally** — because that run fails, v8 never emits the table. CI is the authority.

## Follow-ups found but deliberately not fixed here

1. **A policy with two `webhook` evaluators POSTs to the first URL twice and never calls the second.** `webhook-evaluator.ts` re-derives its config with `policy.evaluators.find(…)` while `buildEvaluators` constructs one instance per entry.
2. **Manual-resolution TOCTOU.** `buildManualResolutionDiff` resolves the base, then `resolveConflict` re-clones and commits against whatever tip that second clone sees, with no expected-base pin (unlike the merge path's `expectedWorkspaceSha`). Pre-existing — but this change turns it into a provenance claim shipped to a third party *and* written to the audit log.
3. **`package-lock.json` is stale at `0.1.0`.** The v0.2.0 release bumped `package.json` without regenerating the lock. Reverted here to keep this PR focused.

https://claude.ai/code/session_01X2igUHWhQWQ2YVfoBk2x3j


jlamoreaux:
Worked through the three "requires further human verification" items. Two were checkable; both hold. One is genuinely not verifiable from here.

### 1. Manual-resolution base provenance — ✅ correct, and now tested

The hypothesis was that `buildManualResolutionDiff` might resolve its base separately from the content reads that build the diff. It does not — and it is structurally *tighter* than `getDiffBetweenRepos`, because there is only one clone to keep consistent:

```ts
const cloneResult = await cloneRepo(projectRemote, projectToken, logger, { ref: branch });
const { fs, dir } = cloneResult.data;                        // one clone
const baseOid = (await fromPromise(git.resolveRef({ fs, dir, ref: branch }))).data;
for (const { file, content } of resolutions) {
  const readResult = await readFileAtCommit(fs, baseOid, file, logger);   // same fs, that oid
  ...
}
return ok({ diff: buildUnifiedDiff(baseContent, resolvedContent), baseSha: baseOid });
```

Every base read is issued against the returned `baseOid` on the same `fs`. There is no second read to disagree with.

**The observation about the test was fair, though**, and it found a real gap: `buildManualResolutionDiff` had **no unmocked coverage at all**, and the route test only asserts the mocked value is forwarded. Fixed in a2ccfda with the real-clone equivalent of the `getDiffBetweenRepos` check — two commits on the base, then assert `baseSha` is the tip *and* the diff runs `v2 -> resolved`, so the base content demonstrably came from the commit `baseSha` names.

### 2. Other entry points calling `runEvaluation` without a context — ✅ none exist

Exhaustive search of `src/`. There are exactly three call sites, all passing a context:

| Call site | Context |
|---|---|
| `services/change-flow.ts:403` | `{ baseSha: diffBaseSha }` |
| `routes/changes.ts:1380` | `{ baseSha: diffBaseSha }` |
| `routes/sync-management.ts:666` | `{ baseSha }` |

No queue consumer and no GitHub-sync path calls `runEvaluation`. The two remaining front doors — `routes/git-http.ts:899` (gated `git push`) and `routes/changes.ts:286` (REST create) — both route through `createChangeWithEvaluation`, i.e. the `change-flow.ts:403` site. And the only `.evaluate(` call sites in `src/` are `change-flow.ts:202` and `composite-evaluator.ts:21`/`:47`, all of which forward the context. So "always present in practice" holds as documented.

### 3. Receiver tolerance — cannot be verified from the repo

Correct, and deliberately left as an operator action: whether a given deployment's receiver rejects unknown fields is a property of code that does not live here. It is called out in the `CHANGELOG.md` entry and in a blockquote in `docs/user-guide/ci-integration.md`, both saying to allow unknown fields *before* upgrading. That is the most this PR can do; notifying existing webhook users is a release-time step, not a code change.

### On the "non-compliant requirement"

Flagging the departure is right, and it is intentional — the issue's literal suggestion (forward `change-flow.ts`'s `resolveProjectHead` value) sends a base the diff may not have been computed against, which replaces an honest absence with a confident wrong answer. It is guarded rather than assumed: implementing the suggested version fails the regression test with `expected 'sha_base' to be 'diff_base_sha'`.

---

Gates after a2ccfda: typecheck clean, biome clean, **2298 passing** (+1). The 43 failures are the pre-existing `node:sqlite` numbered-parameter bug on Node 22.14 ([#340](https://github.com/stratum-eng/stratum/issues/340)), unchanged from `main` and absent on CI's 22.23.2.

jlamoreaux:
Closing as superseded — #319 already shipped this exact feature and has been merged to main. `EvaluationContext.baseSha` is now live end-to-end (`src/evaluation/types.ts`, `webhook-evaluator.ts`, `change-flow.ts`), threaded through the webhook evaluator payload so a receiver can check out the right revision, addressing #274.

This PR's version is an independent implementation of the same interface/field, which is why it now conflicts across 7 files including a duplicate `EvaluationContext` interface declaration. Nothing here adds functionality beyond what already merged, so closing rather than reconciling two competing implementations.

PR #339, opened 2026-08-30
Bound sandbox evaluation with a total budget and ignore install scripts

### LINKED ISSUE #239: Sandbox evaluator: request-path timeout budget and npm lifecycle-script policy
## Context

Found during CodeRabbit review of #211 (issue #194, sandbox evaluator materializes the full evaluated tree). Both items below are real but need a policy/architecture decision rather than a quick fix, so filing as a scoped follow-up instead of folding into #211.

## 1. Install + test can exceed the synchronous request's practical timeout

`SandboxEvaluator.evaluate()` runs inside the request path: `POST /projects/:name/changes` (`createChangeWithEvaluation`) and `POST /changes/:id/evaluate` (`runEvaluation`) both `await` it directly. With defaults, `installTimeoutMs` (120s) + `timeoutMs` (60s) alone allow up to 180s, before counting the repo clone and tree read that happen first. Callers and any intermediate proxy will likely time out before the response arrives, leaving the change row in an intermediate state.

Options to evaluate:
- Bound the total sandbox budget (install + test) to something safely under typical request/proxy timeouts, and fail closed if exceeded.
- Move sandbox execution off the synchronous request path (e.g. queue it and report the result asynchronously), similar to how other longer-running evaluators might already be structured.

## 2. `npm ci`/`npm install` execute untrusted lifecycle scripts

The evaluated tree is authored by an agent or a user who can push to a workspace. Its `package.json` can declare `preinstall`/`install`/`postinstall` scripts, which the install step runs — with network access — before any human review. The Cloudflare Sandbox binding is the intended isolation boundary here, so this isn't necessarily a bypass, but the threat model should be stated explicitly rather than assumed:

- Document what egress/filesystem limits the Sandbox binding actually enforces.
- Consider making lifecycle scripts opt-in via policy (e.g. an `allowInstallScripts` flag that adds `--ignore-scripts` by default) once the threat model is confirmed.

(`--no-audit --no-fund` were already added directly in #211 since they're a pure win with no behavioral tradeoff — this issue is only about the `--ignore-scripts` default, which needs a decision.)

## Source

CodeRabbit findings on PR #211: https://github.com/stratum-eng/stratum/pull/211#pullrequestreview-4978031213

jlamoreaux:
Closes #239.

## Why

`SandboxEvaluator.evaluate()` runs on the synchronous request path — `POST /projects/:name/changes` (`src/routes/changes.ts:286`), the re-evaluate route (`:1376`), and the gated `git push` receive-pack handler (`src/routes/git-http.ts:899`). Two problems, both flagged in #239 as needing a policy call rather than a quick fix:

1. **Nothing bounded total evaluation time.** `installTimeoutMs` and `timeoutMs` were independent, so the old defaults alone permitted 180s of sandbox work before an unbounded tree read — and *before* the three repo clones the request has already done. Both values came straight from `.stratum/policy.yaml` with no validation, so `installTimeoutMs: 999999999` was accepted verbatim.
2. **`npm ci`/`npm install` executed untrusted lifecycle scripts.** The evaluated tree is authored by an agent or anyone who can push to a workspace; its `postinstall` ran before any human review.

Issue #239 offered a bounded budget or moving evaluation onto a queue. This takes the **bounded budget**; async evaluation stays tracked in `docs/REMAINING_WORK.md`, and the budget remains correct inside a future queue consumer.

## What changed

- **`totalBudgetMs`** (default 150s) bounds the whole evaluation. Each phase gets `min(configured, remaining)`; exhaustion returns a *verdict* with a stable, greppable reason naming the phase (`sandbox budget exceeded (install)`) instead of hanging. Default install timeout drops **120s → 90s** so the per-phase defaults sum to exactly the budget — an invariant test enforces it, so a future bump can't silently make truncation the norm.
- **`--ignore-scripts` by default**, with `allowInstallScripts: true` as the opt-in. `.stratum/policy.yaml` is already in `PROTECTED_CONFIG_FILES`, so flipping the flag requires human approval and can't be force-merged.
- **Per-evaluator policy validation** in `policy-loader.ts`, which previously only spread parsed YAML onto the defaults: clamp `sandbox`/`webhook` timeouts and `minScore`, reject a `command` that is blank, over-length, or contains a newline (a shell reads that as a second command), drop malformed `evaluators` entries that used to crash `buildEvaluators` on `.type`, and drop a `webhook` entry with no `url`.
- **Contained `finally` teardown** — a rejecting `sb.destroy()` could replace the returned `Result` and reject `runEvaluation`'s `Promise.all`, discarding every other evaluator's result.
- **[ADR 007](docs/adr/007-sandbox-evaluator-threat-model.md)** states what the Sandbox binding is and is not relied upon to isolate.

## ⚠️ Breaking

Repos whose build needs lifecycle scripts (native modules, a `prepare` step) must set `allowInstallScripts: true`. The usual symptom is **not** a failing install — a native module installs unbuilt and then fails when the test command loads it. No deployed project is affected today: the `[[sandboxes]]` binding is commented out at `wrangler.toml:121-123`, which is why this is the right moment to change the default.

Re-evaluating an existing change now runs under the new default, so a change that passed before may fail on re-evaluation.

## Two security fixes found in review

Both are independent of #239 and worth reviewing on their own:

- **A rejected `minScore` left the raw value in place.** The spread put the unvalidated value in `policy.minScore` before validation, and rejection only skipped the assignment. `minScore: -.inf` (or the string `"-5"`) made `score >= minScore` true for *every* score — accepting changes whose evaluators had all scored 0.
- **A YAML alias cycle silently discarded merge protection.** `JSON.stringify` throws on a circular structure, and it was reached while *logging* a rejected entry. The throw escaped to the handler that treats a file as absent, so a policy declaring `requiredApprovals: 2` became the permissive default with no `configError` — dropping the approval requirement and re-enabling force-merge.

## Honest limits (all in ADR 007)

- The budget bounds time **awaiting the sandbox**, which is the #239 overrun. It does **not** bound Worker-side CPU: Workers freeze `Date.now()` across pure-CPU spans (`src/utils/phase-timer.ts:5-16`), so pack decompression or base64-encoding a large tree is constrained by workerd's CPU limit instead.
- It is **not a request-level bound** — three pre-evaluation clones, `sandbox.create()`, and the timeout-less `llm` evaluator all sit outside it.
- **`--ignore-scripts` is not containment.** A tree-supplied `.npmrc` still redirects the registry, lockfile `resolved` URLs are still fetched verbatim, and the scored command runs untrusted code by design.
- Whether the binding's `timeout` actually terminates a process is undocumented and untestable while the binding is disabled.
- The post-merge smoke check (`src/merge/post-merge.ts`) is a **worse** exposure — write-scoped token, unbounded decode timeout, no install at all. Verified out of scope for code here (it imports only `materializeTree`, never `installCommandFor`), but documented in the ADR as follow-up.

## Also included

`package-lock.json` still said `0.1.0` and lacked the `engines` block that `package.json` gained in the v0.2.0 release, so `npm ci` would reject it. `npm install` synced it. Unrelated to #239 — happy to split it out if you'd rather.

## Verification

| Gate | Result |
|---|---|
| `npm run typecheck` | clean |
| `npm run lint` | zero errors, zero warnings |
| `npm test` | 129 passing across the two touched suites |
| coverage | `sandbox-evaluator.ts` 100% stmts / 89% branches, `limits.ts` 100%, `policy-loader.ts` 83%/80% (floor: 38/72) |
| `npm run release:check` | CHANGELOG well-formed |

`tests/issues.test.ts` and `tests/issues-comment-body-limit.test.ts` have **43 failures that reproduce identically on `main`** (confirmed by stashing this change) — pre-existing and unrelated, but they do mean `npm run test:coverage` aborts before printing its summary, so the coverage numbers above were measured per-file.

https://claude.ai/code/session_01UXq4Zc9xqK3ETqVZ8tLeo2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   off the single unnamed key accounts were given before scoped tokens existed.
   Named tokens are unaffected. Move anything still using the legacy credential
   onto a named token first — disabling cannot be undone.
+- **A total time budget for sandbox evaluation** (`totalBudgetMs`, default 150s). The per-phase
+  timeouts were independent, so nothing bounded their sum and an evaluation could hold a
+  synchronous request open past any client or proxy deadline. Each phase is now granted
+  `min(configured, budget remaining)`, and exhausting the budget returns a failing verdict whose
+  reason names the phase (`sandbox budget exceeded (install)`) instead of hanging. See
+  [ADR 007](docs/adr/007-sandbox-evaluator-threat-model.md), which also documents what the
+  Cloudflare Sandbox binding is and is not relied upon to isolate.
+- `allowInstallScripts` on the `sandbox` evaluator config.
+
+### Breaking
+- **The sandbox evaluator no longer runs npm lifecycle scripts.** Dependency installs now pass
+  `--ignore-scripts`, because the evaluated tree is untrusted and a `preinstall`/`postinstall`
+  would otherwise execute before any human review. Projects whose build genuinely needs them
+  (native modules, a `prepare` step) opt back in with `allowInstallScripts: true` on the
+  `sandbox` evaluator in `.stratum/policy.yaml`. Note the usual symptom is *not* a failing
+  install: a native module installs unbuilt and then fails when the test command loads it.
+- **The default sandbox install timeout drops from 120s to 90s**, so the per-phase defaults sum
+  to exactly the new total budget and an unconfigured project is never truncated.
+- Re-evaluating an existing change (`POST /changes/:id/evaluate`) runs under the new default, so
+  a change that passed before this release may fail on re-evaluation.
 
 ### Fixed
 - Approvals are dismissed when the evaluated **base** moves, not only when the
@@ -105,6 +125,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   takes the project id rather than its name, so a name-scoped lookup is not
   expressible at the call site, and `createWebhook` requires `projectId` so no
   new unstamped row can undo the backfill.
+- Policy files are now validated per evaluator, not just spread onto the defaults. `sandbox` and
+  `webhook` timeouts are clamped into range, `sandbox.command` is rejected if it is blank,
+  over-length, or contains a newline (which a shell would read as a second command), a `webhook`
+  entry without a `url` is dropped, and malformed `evaluators` entries are dropped instead of
+  crashing evaluator construction. An out-of-range value is clamped with a warning; it does not
+  block merges.
+- **`minScore` is clamped to `[0, 1]` and replaced when rejected.** Previously the raw value from
+  the policy file survived validation, so `minScore: -.inf` (or the string `"-5"`) made
+  `score >= minScore` true for every score — accepting changes whose evaluators had all scored 0.
+- A policy containing an unusable evaluator entry now fails the merge gate closed with a
+  `configError` rather than dropping that entry and proceeding on the survivors. Silently removing
+  one gate while its siblings remain would let a change through on the rest — a `webhook` whose
+  `url` was typo'd previously reached the evaluator and blocked. An unrecognised evaluator *type*
+  is unaffected; it is still passed through and rejected downstream.
+- A YAML alias cycle in a policy file no longer causes the whole file to be treated as absent.
+  Serializing a rejected entry for a log line could throw, and the fallback silently discarded the
+  file's `merge` protection — dropping `requiredApprovals` and re-enabling force-merge.
 - Sandbox evaluation of a pinned commit no longer clones a workspace's entire reachable
   history into memory: `readRepoFiles` now clones shallow and grows the fetch window only
   as far as needed to reach the pinned commit, capped at 500 commits, instead of an
--- a/docs/README.md
+++ b/docs/README.md
@@ -47,7 +47,8 @@ docs/
 │   ├── 003-d1-for-import-state.md
 │   ├── 004-high-frequency-agent-commits.md
 │   ├── 005-git-smart-http-proxy.md
-│   └── 006-ssh-transport.md
+│   ├── 006-ssh-transport.md
+│   └── 007-sandbox-evaluator-threat-model.md
 ├── research/                           # Research notes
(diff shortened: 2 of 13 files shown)


jlamoreaux:
Thanks — all three focus areas were real. Fixed in bcdee92.

**1. Any exception during a budget-capped phase became a verdict.** Correct, and the narrow window you describe is the bad one: a transient Sandbox error 200ms into a 5s capped grant was reported as a definitive score-0 judgement about the change.

Classification now requires **both** that the budget shortened the grant *and* that the phase actually spent it:

```ts
const timedOut = phaseConsumedItsGrant();   // elapsed >= the grant handed out
...
if (grantWasBudgetCapped && timedOut) return budgetVerdict(currentPhase);
```

I used elapsed-vs-grant rather than inspecting the error, since matching on message text is binding-specific and would rot silently. New test covers exactly your scenario — capped grant, early throw, still `err`.

**2. The decode step bypassed `grantFor`/`phaseStartedAt`.** Also correct, including the secondary effect on `sandbox_ms`. `materializeTree` now takes the caller's *grant function* rather than a `Budget`, so the decode's grant flows through the same bookkeeping as every other phase and is classified and charged identically. The post-merge caller passes none and is unaffected.

**3. Fail-open shift on partially-dropped evaluators.** This was the most important one — and it's a regression I introduced, since dropping a `url`-less webhook is new in this PR. You're right that the empty-array check cites exactly the reasoning that applies to a partial drop.

Any dropped entry now fails the gate closed, not just the all-dropped case. An unrecognised evaluator *type* is deliberately unaffected: those are copied through and rejected by `buildEvaluators`, so a policy naming a future evaluator doesn't trip it. Three tests pin the boundary.

One consequence worth noting: a YAML alias cycle now yields a `configError` rather than loading. Still the safe direction — before this PR that case silently discarded the file's `merge` protection with no `configError` at all.

Two items from your ticket-compliance notes that I'd flag as genuinely open, not resolved here:

- **The 150s default is a judgement, not a measurement.** No request timeout is configured (`wrangler.toml` has no `[limits]` block) and I did not measure workerd's actual ceiling. It's a named, documented, overridable constant — worth revisiting against a real limit.
- **The Sandbox binding's guarantees are unverified** and can't be tested while `[[sandboxes]]` stays commented out. ADR 007 states them as assumptions rather than facts, deliberately.

PR #358, opened 2026-09-04
Authenticate the CLI through the browser with `stratum login`

jlamoreaux:
## Why

`stratum login` accepted only `--key`, and getting that token was an out-of-band errand: open the settings UI, mint a token, paste it into a terminal. The guard that forces this is deliberate — token management requires a browser session, so a token can never mint another token — but it left the CLI unable to log you in.

**The server already had everything needed.** Stratum runs an OAuth 2.1 authorization server for MCP clients (RFC 7591 dynamic registration, mandatory PKCE, `authorization_code` + `refresh_token`); #356 added loopback redirects with RFC 8252 §7.3 port relaxation, which exists precisely for native apps binding an ephemeral port; and `src/middleware/auth.ts` resolves OAuth access tokens as a peer of API tokens on every route, by explicit design — *"the MCP endpoint reuses the real API handlers, so an OAuth credential has to authenticate everywhere a user token does, or the tools would diverge from the API they wrap."*

The CLI simply never asked. **This is the client half only — zero server changes.**

## What

```bash
stratum login              # opens your browser; nothing to create or paste
stratum login --read-only  # narrower grant
stratum logout             # revokes, then clears
```

The **granted** scope — not the requested one — is printed at login and by `stratum status`, so a narrowed grant is visible immediately rather than at the first write that 403s. That was the `--key` path's quiet failure mode: it validates against `/health`, which authenticates but does not report scope.

`--key` and `STRATUM_HOST`/`STRATUM_API_KEY` still work for CI. One breaking change, in the CHANGELOG: `--host` *without* `--key` now opens a browser instead of prompting, and without a TTY the command refuses rather than hanging.

## The delicate part

Refresh tokens rotate, and the server treats a retired one as theft — presenting it revokes the **entire** grant (`storage/oauth.ts:765`, OAuth 2.1 §4.3.1). Adversarial review found three defects here, each reproduced before being fixed:

| Defect | Consequence |
|---|---|
| `mode` on `writeFile` is honoured only when the call *creates* the file | Everyone upgrading kept a world-readable refresh token at `0644` |
| Two concurrent commands both held the same refresh token | The loser's replay revoked the grant — logged out of every terminal |
| A junk `expires_in` threw *after* the server had retired our token | Rotated token never persisted; silent logout reading `Invalid time value` |

Fixes: writes go through a temp file + `rename` (which replaces the inode and carries `0600` with it) and are fsynced; rotation takes a cross-process lock with an ownership nonce and re-reads the stored credential inside it, aborting rather than rotating if another `login`/`logout` changed it — including one for a *different host*, which would otherwise have sent one host's token to another; and a malformed `expires_in` is coerced rather than thrown.

Also hardened: credentials refused over plaintext and to off-origin endpoints named by a discovery document; timeouts on every OAuth call (a refresh re-enters discovery, so an unreachable host hung *every* command); a stray request to the loopback port is ignored rather than cancelling the login; token fields validated instead of cast (`access_token: 12345` was reachable); only `invalid_grant`/`invalid_client` report as an expired session.

## Verification

- `cli`: **64 tests** (from 11), typecheck, lint — all pass.
- Against **production**: three concurrent commands needing a refresh → one rotation, session intact, no lock or temp files left. Before the fix that exact command revoked the grant.
- Root typecheck passes; `src/` untouched.
- Two gates did not exist and now do: `cli/tests/` was never typechecked (which is what let two `as never` escapes through), and `cli/` was never linted in CI.

## Note for reviewers

`npm test` at the repo root has **43 pre-existing failures on `main`** (`tests/issues.test.ts`, `tests/issues-comment-body-limit.test.ts`). Confirmed by stashing this branch and re-running — identical. Not from this PR, and untouched by it.

Known trade-offs, deliberately not fixed here: null-valued numeric fields crash the `index.ts` output formatters (pre-existing, unrelated to auth); `cli/` still has no coverage floor; `createAgent` is dead code in `client.ts`.

https://claude.ai/code/session_017JSjib4fGRdhYF9AMUazdC

## Summary by CodeRabbit

- **New Features**
  - Added browser-based OAuth login with secure authorization, automatic token refresh, and scope reporting.
  - Added read-only sessions and a `logout` command that revokes OAuth sessions when possible and clears local credentials.
  - Added headless and CI authentication using API keys.
  - Improved authentication errors, request retries, timeouts, and status reporting.

- **Security**
  - Credentials now use restricted permissions, atomic updates, and protection against concurrent access.

- **Documentation**
  - Updated CLI and OAuth guides with login methods, token behavior, scopes, and logout.
  - CLI now requires Node.js 20 or newer.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,10 +47,11 @@ jobs:
         with:
           node-version: "22"
 
-      - name: CLI — install, typecheck, test, build
+      - name: CLI — install, lint, typecheck, test, build
         working-directory: cli
         run: |
           npm install
+          npm run lint
           npm run typecheck
           npm test
           npm run build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,8 +36,14 @@ jobs:
           path: coverage/
 
   packages:
-    name: CLI & Agent Packages
+    name: CLI & Agent Packages (Node ${{ matrix.node }})
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        # The floor each package declares in `engines`, and the version most
+        # contributors run. Testing only the latter lets the floor rot.
+        node: ["20", "22"]
     steps:
       - name: Checkout
         uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
@@ -45,16 +51,19 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with:
-          node-version: "22"
+          node-version: ${{ matrix.node }}
 
-      - name: CLI — install, lint, typecheck, test, build
+      # Lint runs last, per AGENTS.md: Biome autofixes formatting, so putting it
+      # ahead of the correctness gates just reports style on code that may not
+      # compile.
+      - name: CLI — install, typecheck, test, build, lint
         working-directory: cli
         run: |
           npm install
-          npm run lint
           npm run typecheck
           npm test
           npm run build
+          npm run lint
 
       - name: Agent — install, typecheck, test, build
         working-directory: agent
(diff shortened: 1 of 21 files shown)


jlamoreaux:
@coderabbit review

jlamoreaux:
@coderabbit approve

PR #359, opened 2026-09-04
Deploy the merged tree to Cloudflare or Vercel after a merge

jlamoreaux:
## What and why

A project hosted only on Stratum had no way to get merged code onto the internet. `docs/user-guide/ci-integration.md` said so plainly: Stratum executes code in exactly three places, and none of them can publish.

This adds a `deploys:` key to `.stratum/policy.yaml`. After a merge survives the post-merge check, each entry uploads the merged tree to **Cloudflare static assets**, a **Cloudflare Worker script**, or **Vercel**, using credentials from a new encrypted per-project secret store.

## The design decision that shaped everything

The obvious implementation — run `wrangler deploy` in a sandbox with the token in env — is [ruled out by Cloudflare's own guidance](https://developers.cloudflare.com/sandbox/1-0-preview/environment/): *"Do not put live API keys or other long-lived credentials into the sandbox."*

So the untrusted step and the credentialed step are separated instead. Building is untrusted and needs no credential; deploying needs a credential and no untrusted code. The Worker calls the provider HTTP API directly and the secret never leaves it.

Three things fall out of that:

- No CLI is invoked, so there is no `npx` → `node_modules/.bin` hijack path where a repo-supplied `wrangler` runs with the token in env.
- `target: command` was dropped. It was the only in-repo way to echo a secret back out, and the only reason a credential would ever need to reach a sandbox.
- **v1 needs no `SANDBOX` binding.** `[[sandboxes]]` is commented out in every env block in `wrangler.toml`, so a sandbox-based design would have shipped inert. The cost is that v1 has no build step and deploys the tree as committed; `vercel` builds the uploaded source remotely.

## Correctness properties worth reviewing

- **Deploys trigger after `runPostMergeCheck` resolves, not from `change.merged`.** The event is emitted at `changes.ts` *before* the check, and a failed check auto-reverts the merge — so triggering on the event would publish code Stratum had just decided to revert. The integration test runs the real post-merge check against a failing sandbox and asserts **zero** deployment rows.
- **Rows are stamped with merge time, not queue-processing time.** Queues promise no ordering and a retry reorders outright, so processing time would let an older commit land over a newer one. A second guard refuses any deploy whose merge predates an already-succeeded one for the same target.
- **Deploys run on their own `stratum-deploys` queue** with a DLQ and every knob set explicitly, so a slow deploy cannot stall webhook delivery for other tenants sharing the events queue.
- **Retry is refused on non-terminal rows** — otherwise retrying a `pending_approval` row creates a `queued` row and routes around the approval gate.
- **Secrets** are AES-GCM with `(project_id, name)` bound as AAD, the key derived once per deployment, redacted from provider output *before* truncation, and never returned by any route. Both new tables are in the project-deletion cascade.

## Known limitations (all documented in `docs/user-guide/deployments.md`)

- No build step; no preview deploys; no rollback (retry an earlier successful commit); no Netlify.
- The approval gate refuses **agent identities** but accepts a user's scoped API token and MCP OAuth grants — it means "not an agent", not "a human at a keyboard". Stated as such in code and docs rather than overclaimed.
- A `vercel` deploy recorded as `succeeded` means *accepted for build*; Stratum does not poll. (Same shape as GitHub's Deployments API, except GitHub keeps updating status until terminal.)
- **A deploy-only user must still write an `evaluators:` block**, since a policy file without one is treated as malformed and blocks all merges. Prominent in the docs; fixing it means changing the merge gate, which is out of scope here.
- `merge-batch` runs neither the post-merge check nor a deploy — a pre-existing asymmetry, now documented. Fails closed.
- Rotating `DEPLOY_SECRET_KEY` makes every stored secret undecryptable, with no re-encryption path.

`target: cloudflare-pages` is implemented against **Workers Static Assets**, not Pages Direct Upload: the Pages asset endpoints are not in Cloudflare's public API reference, and Cloudflare steers new projects to Workers. Recorded in a comment at the top of the target.

## Verification

- `npm run typecheck` clean · `npm run lint` clean · `npm run release:check` clean
- `npm test`: 3435 passed, 43 failed — all 43 pre-existing in `tests/issues*.test.ts`, confirmed identical at clean `main` in a throwaway worktree
- `npm run check:guides`: all 12 docs mirrors in sync
- Coverage thresholds in `vitest.config.ts` untouched
- ~330 new tests across 10 files, including an integration suite that drives the real queue consumer with a stubbed provider `fetch` and a real D1

## Operator notes

Requires a new `DEPLOY_SECRET_KEY` Wrangler secret, and the `stratum-deploys` / `stratum-deploys-dlq` queues (the staging creation loop in `ci.yml` is updated). The DLQ has no consumer — dead-lettered deploys sit there for manual inspection.

## Summary by CodeRabbit

- **New Features**
  - Added post-merge deployments to Cloudflare Pages, Cloudflare Workers, and Vercel.
  - Added deployment history, status details, approval, retry, provider links, and redacted logs.
  - Added encrypted, project-scoped deployment secret management.
  - Added deployment webhooks for requested, succeeded, and failed events.
  - Added asynchronous queue processing with retries and dead-letter handling.
  - Added validation and clear rejection details for invalid deployment configurations.

- **Documentation**
  - Added deployment guides, API references, configuration details, limitations, and operational requirements.
  - Updated CI integration and FAQ documentation to describe deployment capabilities.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -117,7 +117,7 @@ jobs:
           CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
           CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
         run: |
-          for queue in stratum-events-staging stratum-imports-staging; do
+          for queue in stratum-events-staging stratum-imports-staging stratum-deploys-staging stratum-deploys-dlq-staging; do
             if out=$(npx wrangler queues create "$queue" 2>&1); then
               echo "✓ Created queue $queue"
             elif echo "$out" | grep -qiE "already (exists|taken)|11009"; then
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,9 +68,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   public API reference. Without the optional `CLOUDFLARE_WORKERS_SUBDOMAIN`
   secret a Cloudflare deploy succeeds with no URL. And a project that only wants
   deploys must still write an `evaluators:` block, because a policy file without
-  one is malformed and blocks every merge. Limits (2,000 files, 25 MiB total,
-  10 MiB per file) are enforced before the first provider request, so a deploy
-  never publishes half a tree. Full reference:
+  one is malformed and blocks every merge. Limits (16 `deploys:` entries per
+  policy file, 2,000 files, 25 MiB total, 10 MiB per file) are enforced before
+  the first provider request, so a deploy never publishes half a tree. Full reference:
   [Deployments](docs/user-guide/deployments.md).
 - **`stratum login` signs you in through the browser.** The CLI had one way to
   authenticate — `--key`, with a token you first created by hand in the settings
@@ -238,6 +238,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   grants as well as scoped tokens.
 
 ### Fixed
+- **A post-merge deploy cannot be lost or run twice.** Four fixes to the new
+  deployment path, all invisible until they bite: the deploy queue ran two
+  messages per isolate, and two concurrent Vercel deploys exceeded the isolate's
+  memory before a terminal status could be written, stranding rows at `running`
+  (`max_concurrency` is now 1); the runner had no deadline, so its storage lease
+  could expire mid-upload and let a second consumer publish the same commit (it
+  now gives up strictly before the lease can); a queue send that failed silently
+  dropped the deploy — after a merge there was no row to recover at all, and
+  after approve or retry the row sat `queued` forever while the response advised
+  a retry that could not work (the request is now recorded durably and picked up
+  by the existing five-minute sweep); and several rejected `deploys:` entries
+  collapsed into one row, so only one of their reasons survived — a rejection
+  sharing a name with a valid entry could even take that entry's row and stop it
+  deploying. Every rejected entry now gets its own `failed` row.
 - **The header fits a phone.** Signed in, the nav's five links were laid out in
   one unwrappable row, so below about 460px the row ran past the viewport, the
   page scrolled sideways, and the wordmark and "logout" were clipped off either
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,67 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   rather than folded into this change.
 
 ### Added
+- **Stratum can deploy a merged change.** A new `deploys:` block in
+  `.stratum/policy.yaml` names one or more deploys, each with a `target`
+  (`cloudflare-pages`, `cloudflare-workers`, or `vercel`), an optional `dir` to
+  publish, the `secrets` it needs, and an optional `requiresApproval` gate.
+  After a change merges *and* the post-merge check declines to revert it, the
+  merged tree is read at the pinned merge commit — configuration and code are
+  therefore provably the same commit — and published to the provider. Every
+  attempt is a row you can see: status, reason, provider URL, duration, and (for
+  project writers) a redacted log tail, on a new deployments page and at
+  `GET /api/projects/{namespace}/{slug}/deployments`. `deployment.requested`,
+  `deployment.succeeded` and `deployment.failed` are subscribable from a project
+  webhook.
+
+  Credentials live in a new **per-project secret store**, encrypted with AES-GCM
+  under a `DEPLOY_SECRET_KEY` the instance holds as a Wrangler secret, with the
+  project id and secret name bound as additional authenticated data. There is no
+  read path: no API, UI, or CLI surface returns a stored value, and only the
+  deploy runner decrypts one. Managing secrets requires a project admin who is
+  **not** an agent — an agent token is refused on every secret route even when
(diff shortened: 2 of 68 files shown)


jlamoreaux:
Addressed all 15 review findings in 15477c0. Every one was verified against current code before acting; none were skipped as invalid.

**The four correctness findings**

- **Isolate OOM before the terminal write** — chose `max_concurrency = 1` over a Vercel-specific byte cap, which would have to be re-derived whenever any target changes how it buffers, and would reject trees that deploy fine today. The ~91.7 MiB peak is now an asserted constant (`MAX_PEAK_DEPLOY_BYTES`). The body is deliberately *not* streamed: a chunked body may or may not be accepted by Vercel's create-deployment endpoint, and nothing in this repo can establish which — every target test injects `fetch`. Streaming is noted as the prerequisite for ever raising the concurrency.
- **Lease could not expire safely** — chose a 10-minute attempt deadline over longer leases. Raising the leases past the 15-minute consumer wall limit would leave rows `running` for 20+ minutes after the platform had already killed the runner. The deadline instead makes "the lease expired" mean "no runner is alive", which is what every reclaim decision already assumed, and it aborts the provider fetch rather than leaving it racing behind a resolved promise.
- **Failed `queue.send` stranded work** — a deployments sweep cannot fix the merge-path case, because there is no row to sweep. All three send sites now record to the existing event outbox, which already has both halves (durable D1 row + five-minute sweep). The outbox row uses an internal type kept **outside** `StratumEvent` and routed to a forwarding-only handler, so `*`-subscribed webhooks never see a work record and the deploy still never runs on the events queue. Residual gap, worth a follow-up: this fires when the failure is *observed*; an isolate that dies between the status write and the send still leaves a `queued` row that only a stale-queued sweep would recover.
- **Rejected entries collapsed into one row** — fixed, and it surfaced a sharper bug in passing: a duplicate-name rejection was taking the unique-index key its *accepted* sibling needed, so that deploy row was never created and never ran.

**On the entry cap** — capped at 16, but reported as a *single* rejection rather than one per excess entry, so a large policy file cannot flood the `deployments` table.

**On the 401/403 split** — confirmed intentional and now documented with a caller×route table: `requireSecretsAdmin` checks `agentId` first and returns 403 ("we know who you are, this isn't yours"), while approve only checks `userId`, which an agent token never sets, giving 401 ("we don't know you as a *user*").

One behavior change reviewers should note: **approve and retry now answer 200/201 where a failed enqueue previously produced a 500**, because the request is durable. A 500 there now means the outbox write failed too. Documented in `docs/api/endpoints/deployments.md`.

Gates: typecheck clean, lint clean, `release:check` clean, 12 docs mirrors in sync, 3469 tests passing (the 43 failures in `tests/issues*.test.ts` are pre-existing on `main` — verified in a clean worktree).

jlamoreaux:
Fixed in e4309bb — this was a good catch and more reachable than it looks.

`resolveTimestamp` validated ISO 8601 and returned the string **unchanged**, while the accepted grammar admits three mutually incomparable spellings of the same instant (`Z`, a `±HH:MM` offset, and an optional fractional part). Since SQLite compares these columns as TEXT, lexical order stopped matching chronological order in both directions:

- `2026-09-04T17:20:43-04:00` sorts **before** `2026-09-04T21:20:43.000Z` — same instant, `17` < `21`
- `2026-09-04T21:20:43Z` sorts **after** `2026-09-04T21:20:43.000Z` — `.` is 0x2E, `Z` is 0x5A

What made it reachable rather than theoretical: `created_at` now carries the *merge time*, which arrives over the queue in `DeployQueueMessage.mergedAt`, so it is not necessarily produced by the same code path that writes the other rows. The columns were comparable only by convention.

`resolveTimestamp` now returns `new Date(Date.parse(value)).toISOString()`. Every write and comparison in the module already funnels through it, so one change covers `insertDeployment`, `claimDeployment`, `completeDeployment`, `supersedeOlder`, `findNewerSucceededDeployment` and `approveDeployment`. `lease_expires_at` was already canonical (derived via `toISOString`) — the vulnerable side of that comparison was the caller-supplied clock, which is what the tests drive.

**On the suggested regression pairings:** the `-04:00` older row vs `Z` newer row case is actually benign, because a negative offset always sorts the string low, which is the direction `supersedeOlder`'s `<=` and `findNewerSucceededDeployment`'s `>` already want. The corrupting cases are the mirrors — an offset-form row that is genuinely *newer* than the keeper, and a newer succeeded row in offset form that is invisible to a `Z` candidate. Covered both orientations rather than only the named one.

Ten cases added; **nine fail if the normalization is reverted** (the tenth is a false-positive guard that correctly passes either way). No existing test depended on the verbatim string — every timestamp literal reaching deployments storage was already canonical, so this is a no-op for them.

I left `isIsoTimestamp` validating without rewriting: a `value is string` guard cannot return a rewritten value, and a message in transit is never compared against a column, so normalizing there would add a second place that has to agree on the canonical form without fixing anything.

Gates: typecheck clean, lint clean, 3479 passing, integration 36/36. (The 43 failures in `tests/issues*.test.ts` are pre-existing on `main`.)

jlamoreaux:
@coderabbitai review

All findings from the first review (`892b891`) are addressed in `15477c0`, and the timestamp-canonicalization finding from the second pass is addressed in `e4309bb` — which landed just after that review was submitted, so it hasn't been looked at yet.

Requesting a fresh pass over the current head so the `CHANGES_REQUESTED` verdict from the first review can be re-evaluated against the code as it now stands.

jlamoreaux:
@coderabbit approve

PR #361, opened 2026-09-04
Create the production queues before deploying to production

jlamoreaux:
## The failure

The first push to main after #359 deployed staging and then failed production:

```
✘ [ERROR] Queue "stratum-deploys" does not exist.
```

[Run 33927843126](https://github.com/stratum-eng/stratum/actions/runs/33927843126). **Production is currently un-deployable** until this lands (or the queues are created by hand).

## Why staging survived and production didn't

#359 added two production queues, `stratum-deploys` and `stratum-deploys-dlq`, and `wrangler deploy` fails outright on a bound queue that does not exist.

Staging already had an `Ensure Staging Queues Exist` step, so it created them and deployed fine. Production never had one — its original queues predated CI, so nothing ever created a queue on that path and the gap stayed invisible until a new queue was bound. #359 added the staging entries to an existing step and, reasonably but wrongly, assumed production had the same safety net.

## The fix

Adds the equivalent step to **both** production deploy paths:

- `ci.yml` — the automatic deploy on push to main
- `deploy-production.yml` — the manual redeploy, which hits the same wall and is exactly what someone reaches for when production is already unhappy

Both loops list **every** production queue rather than only the two new ones, so binding the next queue cannot reopen the same hole. Creation is idempotent — an "already exists" response counts as success — and any other failure still stops the deploy rather than proceeding to a broken release.

## Note on ordering

The step runs after D1 migrations and before `wrangler deploy`, mirroring staging. Verified both workflows parse and the step lands in the right position in all three jobs.

PR previews are unaffected: `pr-preview.yml` generates its own `wrangler.preview.toml`, which binds no queues.

## Summary by CodeRabbit

- **Chores**
  - Production deployments now automatically provision the required message queues before deployment.
  - Existing queues are handled safely without interrupting the deployment process.
  - Unexpected queue-provisioning errors now stop the workflow, helping prevent incomplete or misconfigured production deployments.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,14 +120,19 @@ jobs:
           npx wrangler d1 migrations apply stratum-staging --remote --env=staging
           echo "✓ Migrations applied successfully"
 
-      # Staging binds its own -staging queues (see wrangler.toml); a deploy
-      # fails if a bound queue doesn't exist, so create them idempotently.
+      # A deploy fails outright on a bound queue that does not exist. The names
+      # come from wrangler.toml rather than a copy kept here, so binding a new
+      # queue cannot leave a deploy path un-provisioned.
       - name: Ensure Staging Queues Exist
         env:
           CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
           CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
         run: |
-          for queue in stratum-events-staging stratum-imports-staging stratum-deploys-staging stratum-deploys-dlq-staging; do
+          node scripts/wrangler-queues.mjs staging > /tmp/queues-staging.txt
+          # Redirect, not a pipe: a `while ... | read` loop runs in a subshell,
+          # where `exit 1` would end the subshell and let the step pass.
+          while read -r queue; do
+            [ -z "$queue" ] && continue
             if out=$(npx wrangler queues create "$queue" 2>&1); then
               echo "✓ Created queue $queue"
             elif echo "$out" | grep -qiE "already (exists|taken)|11009"; then
@@ -136,7 +141,7 @@ jobs:
               echo "$out"
               exit 1
             fi
-          done
+          done < /tmp/queues-staging.txt
 
       - name: Deploy to Cloudflare Staging
         id: deploy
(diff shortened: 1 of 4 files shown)


jlamoreaux:
Good catch — addressed in d05f4c3, and it was sharper than the "Trivial" label suggests.

The first version kept its own copy of the queue names, which is the *same* failure this PR exists to fix (a queue bound in config that no deploy path provisions) just relocated rather than removed. Binding a new queue in `wrangler.toml` and forgetting the workflow would have broken the next deploy exactly as it did this time.

`scripts/wrangler-queues.mjs` now reads the names out of `wrangler.toml` for a given environment, and all three provisioning steps consume it — `ci.yml` staging and production, plus the manual `deploy-production.yml` redeploy. There is no second list to update.

Details worth flagging for review:

- **The DLQ is only named by a consumer's `dead_letter_queue`**, never by a block of its own, so a reader that matched `queue = ` alone would silently miss `stratum-deploys-dlq` — the exact queue at issue here. Covered by a test.
- **Hand-rolled rather than a library.** No TOML parser is on the dependency list, and pulling one in to read a file this repo controls is a worse trade than ~30 lines that understand section headers and `key = "value"`.
- **An unknown env name exits non-zero** instead of printing nothing. Silence would mean "provision nothing" and the deploy right after would fail on the first bound queue.
- **The loops read from a file, not a pipe.** `while ... | read` runs in a subshell, where `exit 1` ends the subshell and lets the step pass — the provisioning failure would have been swallowed.

`tests/wrangler-queues.test.ts` asserts the reader sees everything the file binds: the DLQ is found, staging and production names never mix (they are account-level, and sharing them once delivered staging's messages to the production consumer), and **every** queue name in the file is attributed to some environment — so a new `[[env.*.queues.*]]` block that CI never provisions fails the test rather than the deploy.

Gates: typecheck clean, lint clean, 3483 passing, both workflows parse with the step in the right position in all three jobs. (The 43 failures in `tests/issues*.test.ts` are pre-existing on `main`.)

---

Separately, zizmor flagged **overly broad permissions** on `ci.yml` (no `permissions:` block, lines 1-293). That is pre-existing and repo-wide rather than introduced here, and production is currently un-deployable — so I have deliberately left it out of this hotfix rather than change workflow permissions under time pressure. Worth its own PR across all workflows.

jlamoreaux:
@coderabbit approve

PR #365, opened 2026-09-04
Make docs/ the single source, then correct what had drifted

jlamoreaux:
Two halves: make every published page traceable to one file, then fix what the absence of that guarantee had allowed to rot.

Addresses all four review findings from the earlier PR-Agent pass — see the bottom.

## Part 1 — one source per page

`reference/endpoints.md` was **hand-authored on the site** while `docs/api/endpoints/` held 578 lines describing the same API that **nothing published**. Neither was generated from the other and `check:guides` covered neither.

- `/reference/endpoints/` is now **generated** from `docs/api/endpoints/` (README intro + each resource file, headings demoted fence-aware, sibling links → in-page anchors). Same URL, same one-overview structure.
- `check:guides` now also asserts **every published page is either generated or declared in `SITE_OWNED` with a reason**, and every file in `docs/api/endpoints/` is registered in `ENDPOINT_ORDER`.

The second bullet is the part that helps later. The freshness check could only ask *"are the pages I know about stale?"* — structurally unable to see a page nobody generates. That blind spot is how this happened.

**Verified by breaking it:** an unmanaged page and an unregistered endpoint file each fail with `exit 1` and a message naming the fix.

### Content recovery first

Regenerating as-is would have **deleted real content** — I measured 59 prose lines and one route. Worst case: the site documented `DELETE /api/agents/{id}` as *"the only way to revoke"* a credential that **never expires**, and the canonical file had neither the route nor the fact.

Drift ran **both ways**: `issues.md` was stale in the *canonical* copy, where the site correctly said `PATCH` requires a user identity. Acceptance test now reports **0 routes lost**.

## Part 2 — what the audit found

Every page audited against the code. The headline: **the docs recommend a feature the project itself does not use.**

`[[sandboxes]]` is commented out in **every** environment, so the `sandbox` evaluator doesn't skip — it **fails closed**, and naming it in `requiredEvaluators` **blocks every merge in the project**. `getting-started.md` handed exactly that policy to first-time users; the README example carried the same trap twice. This repo's own `.stratum/policy.yaml` uses only `diff` and `llm` — that's the tell.

Also corrected, each verified against code:

| Claim | Reality |
|---|---|
| A conflicting merge falls back to a squash | Returns `409 MERGE_CONFLICT` with a `conflictId`. Untrue for some time |
| README's workspace-creation path | Route does not exist |
| "`wrangler.toml` declares all of them" | `[[sandboxes]]` is commented out |
| webhook `timeoutMs: 300000` | Silently clamped to the 120000 cap |
| CLI at "full API parity" | No command for deployments, orgs, webhooks, secrets |
| "plus AI Gateway" | Appears nowhere in the repo |
| 5 curl examples in `issues.md` | All missing the `@` every namespace requires — each would 404 |
| Import stall recovery "5 minutes" | 20 minutes, with `queued` excluded |
| OAuth replay "revokes tokens from that code" | Revokes **every grant that client holds for that user** |
| One merge Durable Object | Three; MCP server omitted from capabilities entirely |

`errors.md` gained the deployment codes, `413`, and several returned-but-undocumented codes. Reverse direction checked too — no phantom codes.

### The squash claim is worth a note

It propagated from a **stale docstring** on `mergeWorkspaceIntoProject` into the README, getting-started, and the roadmap. The docstring now states the current behaviour *and why it is worded that way*, so the next reader doesn't restore the old sentence.

### And my own work

The deployments quickstart I shipped this afternoon had **five errors**, including a policy example listing `CLOUDFLARE_WORKERS_SUBDOMAIN` under `secrets:` — which makes it **required**, so the example contradicted the paragraph above it and would fail the deploy it was teaching.

## Earlier review findings

| Finding | Resolution |
|---|---|
| Dropped endpoint docs | Root-caused: canonical files were thinner. Reconciled; 0 routes lost |
| Broken anchor rewrite (`#a#b`) | A link with a fragment now keeps the fragment; the heading is on the same page |
| Dead edit link (directory 404s) | Points at `docs/api/endpoints/README.md` |
| Unused `fileURLToPath` | Removed, with the unused exports it was left over from |

## Verification

`check:guides` 13 mirrors in sync · site builds (17 pages) · typecheck clean · lint clean · `release:check` clean · 3483 tests passing (the 43 `tests/issues*.test.ts` failures are pre-existing on `main`).

The one `src/` change is a comment.

## Summary by CodeRabbit

- **Documentation**
  - Expanded API references for authentication, agents, organizations, projects, deployments, issues, errors, and merge conflicts.
  - Clarified OAuth lifetimes, agent permissions, credential restrictions, development login, MCP authorization, and deployment secrets.
  - Documented Cloudflare Sandboxes beta availability, fail-closed evaluators, post-merge checks, CI setup, and troubleshooting.
  - Improved website navigation, endpoint organization, and documentation consistency.

- **Behavior Clarifications**
  - Merge conflicts now return structured conflict details instead of falling back to squash merges.
  - Deployment retries remain supported while restricted agent actions are documented.
  - Import recovery targets actively progressing stalled imports.


jlamoreaux:
Good flag, and the reasoning is sound — but the path is correct as written. Verified against four independent sources rather than the route definition alone, since "this looks like a typo" deserved more than one confirmation:

| Source | Evidence |
|---|---|
| Route definition | `src/index.ts:187` mounts `workspacesRouter` at `/api/workspaces`; `src/routes/workspaces.ts:52` declares `app.post("/:namespace/:slug/workspaces", …)` |
| OpenAPI spec | `docs/api/openapi.yml:2470` — `/api/workspaces/{namespace}/{slug}/workspaces` |
| Shipping CLI | `cli/src/client.ts:256` builds exactly that path |
| Tests | `tests/workspaces-error-taxonomy.test.ts:182,204,226` request `/api/workspaces/@owner/repo/workspaces` |

And the inverse: `projectsRouter` declares **no** workspaces route, so the previous `/api/projects/@you/my-project/workspaces` was a 404. That is what this change fixed — the doubled segment is the real shape, not something introduced while fixing namespace prefixes.

You are right that it breaks the pattern, though. This route is project-scoped but does not live under `/api/projects`, and the very next line in the same example does. That inconsistency is presumably how the README came to document a path that never existed.

So I have added a comment in the example saying the repetition is deliberate and pointing at the mount. A path that looks like a typo and isn't will get "fixed" back otherwise — the same way this PR found the squash-fallback wording had outlived the code and propagated into three documents. Pushed in 24c602e.

Worth noting separately: making the route consistent (`/api/projects/{ns}/{slug}/workspaces`) is a breaking API change with a CLI, a spec and tests behind it, so it belongs in its own change rather than a docs PR.

jlamoreaux:
@coderabbit review

jlamoreaux:
All nine CodeRabbit findings verified against the code first. All nine held, so all nine are fixed in f81fcb6. The branch also now carries a merge of `main`, which clears the conflict.

| Finding | What the code says | Fix |
|---|---|---|
| "silently skipped" (`CURRENT_CAPABILITIES.md`) | `post-merge.ts:40` logs `Post-merge command configured but SANDBOX binding is absent` before returning `status: "skipped"` | "skipped with a warning", and says where the skip is visible |
| `#362` timing row | `8m00s` is 480s against a 420s `config.ai_timeout` — "~10s of margin" could not be right | Column relabelled **Job wall clock** with a paragraph on why it is an upper bound on the call, not the call. The two killed rows read `20m` because that is `timeout-minutes`, which is the tell that the column is job time. Margin claim dropped; per-request durations live in AI Gateway analytics |
| Merge-blocking claim unscoped | `requireAll` defaults to `true` (`composite-evaluator.ts:63`) and the aggregate is `every(passed)`; at `false` it is `some(passed)` | Split into the two cases, and noted that a `merge.requiredEvaluators` entry blocks on its own either way (`protection.ts:59-65`) |
| `deploys:` "works out of the box" | Needs a `target`, provider credentials, `DEPLOY_SECRET_KEY`, and `DEPLOY_QUEUE` | Now "no Sandbox binding, but not zero-configuration", with the prerequisites listed and a link to the deployments guide |
| `merge-batch` in the no-row row | The handler (`changes.ts:898-1253`) calls neither `runPostMergeCheck` nor `enqueueMergeDeploy` | Own row. Approve and Retry both need an existing row and no route creates one, so it is stated as **no recovery**, not a case with a fix |
| 409 always carries paths | The `MergeNotSupportedError` branch (`git-ops.ts:1234-1237`) constructs `MergeConflictError` with an empty `conflictingFiles` | Qualified in both cited places — the docstring and `ROADMAP.md` |
| Duplicate slugs in `ENDPOINT_ORDER` | Set membership passed a repeated slug through both loops | Duplicate check added. Verified by planting a duplicate: exits 1 with `ENDPOINT_ORDER lists "projects" more than once` |
| `publishedPages()` per `SITE_OWNED` entry | Tree walked once per declared entry | Walked once, reused |
| R2 as required | `run-backup.ts:116` warns and skips when `BACKUPS` is unbound; README and FAQ already said optional | Getting-started now matches them |

The merge conflict was in the cost answer in `faq.md`: `main` rewrote it for the relicensing while this branch corrected "AI Gateway" to "Workers AI". Kept both — the evaluator uses the `[ai]` binding (`wrangler.toml:75`); AI Gateway is the PR-review CI path, which is a different thing that happens to share a name.

One correction to this PR's own description: it claimed 43 `tests/issues*.test.ts` failures were pre-existing on `main`. They are gone as of the merge — **3716 tests pass, 0 failures**. Also clean: typecheck, lint, `release:check`, `check:guides` (13 mirrors in sync), site build (17 pages).

---

jlamoreaux:
Round two: four more findings on `f81fcb6`, all verified against the code, all valid, all fixed in `a57433b`.

**Two of them were mine**, introduced when I resolved the conflict with `main` — the cost answer omitted Durable Objects, R2 and Analytics Engine, and scoped Sandboxes to "those evaluators" when `merge.postMergeCommand` needs the binding too. That paragraph now points at the README's own binding table and names what each optional binding is actually for.

**`allowForce` (#3)** is right, and the interesting part is what it does *not* rescue. Force merge bypasses `checkMergeProtection` entirely, so `allowForce: true` does let a writer force other changes past a failed `sandbox` evaluator. But it is deny-by-default (`changes.ts:477-480`), and it is refused outright on a change that touches `.stratum/policy.yaml` (`changes.ts:483-488`) — which is exactly the edit you need to remove the bad evaluator. So the escape hatch does not open on the trap the paragraph is describing. Both halves are now stated.

**Queue-send failures (#1)** — correct, and there is more recovery than the finding assumed. `enqueueMergeDeploy` catches the failed `send` and writes the request to the **event outbox**, which a five-minute cron sweeps and forwards to `DEPLOY_QUEUE`, so the row shows up late rather than never. Only if that outbox write also fails is the request genuinely gone, and the log says exactly that: `Post-merge deploy request could not be made durable and is lost`. The table now distinguishes "appears late" from "gone", since the operator action differs.

One thing I corrected while writing it: my first draft linked the README binding table at `#deployment`. The table lives under `### Prerequisites`, so that anchor would have 404'd. Fixed before the push.

Docs only this round — no `src/` or script changes. Verified: 3716 tests pass, typecheck and lint clean, `check:guides` 13 mirrors in sync, site builds (17 pages).

---

jlamoreaux:
@coderabbitai review

Requesting the pass on `a57433b` now that the rate limit has lifted — the previous run was cut off mid-review, so the last commit reviewed is `f81fcb6`.

---

jlamoreaux:
@coderabbitai review

Retrying `a57433b` — the 00:30 request was rate-limited rather than run, so the last reviewed commit is still `f81fcb6`. Both prior reviews are fully addressed and all 13 threads are resolved; CI is green on `a57433b`.

---

PR #366, opened 2026-09-05
Make PostHog analytics answer product questions

jlamoreaux:
## Summary

Telemetry sent **two** events with almost no properties: `api_request` (route pattern, method, status, latency) and `stratum.<type>` (type, actor type, project id). It could count merges. It could not answer what fraction of changes pass evaluation, what reviewers decide, which deploy targets fail, which MCP tools agents actually call, or how anyone signed up.

Two problems, fixed in that order.

**1. A structural hole.** Export is governed by two gates, but only one — the instance switch — lived in the client. The per-user opt-out (#257) lived at each `capture()` call site, which is how the queue exporter originally shipped ignoring it entirely. `posthog.ts` carried a docblock warning about exactly this, but a comment cannot bind a call site that never reads it, and this PR adds several call sites.

`AnalyticsTracker` (`src/analytics/tracker.ts`) closes it structurally: no public constructor, and every factory must produce an `AnalyticsActor` whose preference is already resolved. A call site cannot reach `capture()` without having decided, and the lookup paths still fail closed. Behaviour at the existing call sites is unchanged.

**2. The data itself.** `src/analytics/events.ts` is the new catalog — one file that answers "what leaves my instance", with the privacy rule stated once: every property is a source-code literal or a bounded value.

### What is now captured

- **Domain events carry their payload** (whitelisted per type): evaluation `score`/`passed`, review `verdict`, deploy `target`, import `source_provider`, and whether an issue closed through a linked change. All of this was already computed and then discarded.
- **`mcp_request`** — one per JSON-RPC message at `/mcp`: method, tool, outcome, and the connecting client's name/version. The flagship agent surface previously emitted nothing.
- **`auth_completed`** — sign-up vs sign-in, by provider. One helper, called next to the `recordAudit` that already marks the same moment across five handlers in three files.
- **`error_occurred`** — an unhandled exception never reaches the analytics middleware's post-`next()` body, so a 500 born that way was invisible entirely.
- **`background_job_completed`** — abandoned outbox events and webhook delivery outcomes.
- **`api_request`** gains `surface` (api/ui/git/mcp/auth/admin/internal) and `actor_type`.
- **Every event** gains `environment`, from a new `STRATUM_ENVIRONMENT` var, so staging stops polluting production's numbers.

### Because this is open source

`docs/user-guide/faq.md` now lists every event and every property exhaustively, and `tests/analytics-catalog.test.ts` **fails the build** if that list and the code disagree. Free text is excluded by construction: exception messages, deployment failure reasons, and import URLs are dropped; MCP tool names are validated against the tools this build defines rather than echoed back; a self-hosted GitLab collapses to `other` rather than reporting a hostname that identifies the company running it.

## Related issues

Builds on the per-user opt-out from #257.

## Type of change

- [ ] Bug fix
- [x] New feature
- [x] Refactor / cleanup
- [x] Documentation
- [ ] Other:

## How was this verified?

- [x] `npm run typecheck`
- [x] `npm test`
- [x] `npm run lint`

56 new tests across four suites (`analytics-catalog`, `analytics-tracker`, `analytics-mcp`, `analytics-auth`) plus additions to `analytics-middleware`. `src/analytics` is at **99.4% statements**; `tracker.ts` and `posthog.ts` at 100%. Global coverage 73.7% statements / 80.7% branches, comfortably above the thresholds.

`cd website && npm run check:guides` reports all 12 mirrors in sync.

> **Note:** 43 tests fail in `tests/issues.test.ts` and `tests/issues-comment-body-limit.test.ts`. These are **pre-existing** — they fail identically on the base commit (`cb04046`) and are untouched by this PR.

## Checklist

- [x] `CHANGELOG.md` updated (`## [Unreleased]`)
- [x] Public docs updated (`docs/user-guide/faq.md`, `docs/developer/deployment.md`, mirrors regenerated)
- [x] No client-side JavaScript added

## Reviewer attention

Three judgment calls worth a second opinion:

1. **Property rename.** `projectId` → `project_id` and `actorType` → `actor_type`, for consistency with `latency_ms` and every new property. This breaks any existing dashboard on those names; the CHANGELOG documents it with a `coalesce()` bridge. Easy to revert if there are dashboards in use.
2. **No CLI or agent telemetry.** `cli/` and `agent/` remain uninstrumented. Adding silent phone-home to a locally installed OSS binary is a different privacy decision from server-side telemetry — no opt-out UI, no account context at install time. Deliberately left as a maintainer call.
3. **`client_name` / `client_version` are the only free-text fields**, capped at 64 characters and documented as such in the FAQ. They are software self-identification, like a User-Agent, and "which editors connect" is worth knowing — but it is the one place the literals-only rule is relaxed.

## Follow-ups not in this PR

- PostHog **groups** for project/org, enabling native per-project adoption counts (`project_id` currently rides as a plain property).
- Per-evaluator breakdown on `change.evaluated`, which needs a new field on the outbox payload.

https://claude.ai/code/session_013nJ5fwv3FwrbuTraoQBHP5

## Summary by CodeRabbit

- **New Features**
  - Expanded analytics coverage for authentication, errors, background jobs, MCP requests, repository activity, and webhook delivery.
  - Added deployment environment labels, timestamps, version metadata, and consistent event properties.
  - Added privacy safeguards, including opt-out enforcement, bounded event data, owner attribution, and personless system metrics.

- **Documentation**
  - Expanded telemetry FAQs with the event catalog, supported properties, identity handling, and privacy limitations.
  - Documented environment configuration for staging and production.

- **Tests**
  - Added comprehensive coverage for analytics, privacy filtering, opt-outs, MCP reporting, and environment attribution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -215,6 +215,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   or free-text message. `docs/user-guide/faq.md` lists every event and every
   property exhaustively, and a test fails the build if that list and the code
   disagree.
+- **An agent no longer counts as a person in analytics.** Every agent token had
+  its own `distinctId`, so PostHog minted a person profile per credential —
+  splitting one human's history across several profiles, inflating the billed
+  person count, and making every user count, funnel, and retention curve wrong
+  by however many agents happened to be running. An agent acts under its
+  owner's account (the owner's opt-out already governs it), so its events are
+  now attributed to that owner, with the agent's own id carried as `agent_id`
+  so the breakdown is not lost.
+- **Domain events are dated from when the activity happened.** They were dated
+  from when the queue consumer exported them, which a retry or the five-minute
+  stale sweep can delay well past the fact — skewing every time series toward
+  whenever the queue drained. The outbox row's own timestamp is now sent.
+- **Every event carries `$lib_version`**, so a change in a metric can be tied to
+  the release that caused it instead of guessed against deploy times.
+- **A person profile now has something on it.** `signup_provider` is recorded
+  once per account, as `$set_once`, so a cohort like "accounts that signed up
+  with GitHub in August" is expressible — previously a person was an opaque id
+  with no properties at all. Email, username, and display name are still never
+  sent.
 - **Analytics property names are now consistently snake_case.** The `stratum.*`
   events sent `projectId` and `actorType`; they now send `project_id` and
   `actor_type`, matching `latency_ms` and every new property. Dashboards built
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -198,6 +198,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **Product analytics now answers questions instead of just counting requests.**
+  Telemetry sent two events with almost no properties: a request event carrying
+  a route pattern, and a repository-activity event carrying only its type, the
+  actor type, and a project id. It could count merges; it could not say what
+  fraction of changes pass evaluation, what reviewers decide, which deploy
+  targets fail, which MCP tools agents actually call, or how anyone signed up.
+  Four new events (`mcp_request`, `auth_completed`, `error_occurred`,
+  `background_job_completed`) and per-type properties on the `stratum.*` events
+  close that. `api_request` gains `surface` and `actor_type`, and every event
+  now carries an `environment` label so staging traffic stops polluting
+  production's numbers — set `STRATUM_ENVIRONMENT` in each `[env.<name>.vars]`
+  block (named environments do not inherit top-level `[vars]`). The privacy
+  rule is unchanged and now enforced in one place: every property is either a
+  source-code literal or a bounded value, never a name, path, URL, title, ref,
+  or free-text message. `docs/user-guide/faq.md` lists every event and every
+  property exhaustively, and a test fails the build if that list and the code
+  disagree.
+- **Analytics property names are now consistently snake_case.** The `stratum.*`
+  events sent `projectId` and `actorType`; they now send `project_id` and
+  `actor_type`, matching `latency_ms` and every new property. Dashboards built
+  on the old names need updating — an insight can bridge the gap with
+  `coalesce(properties.project_id, properties.projectId)`.
+- **The per-account telemetry opt-out is now structurally impossible to skip.**
+  It lived at each `capture()` call site rather than in the client, which is how
+  the queue exporter originally shipped without honouring it at all. A new
+  `AnalyticsTracker` cannot be constructed without a resolved preference, so a
+  call site cannot forget one it never has to know about. Behaviour for existing
+  call sites is unchanged, including failing closed when the preference cannot
+  be read.
 - **`stratum login --host <url>` without `--key` now opens a browser** instead of
   prompting for a key. Scripts that piped a key into the prompt must pass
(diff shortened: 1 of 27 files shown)


jlamoreaux:
Thanks — worked through all seven. Four were real, and three of those were the same class of mistake on my part: a docblock asserting a guarantee the code did not actually enforce. Fixed in bcab746.

**Fixed**

- **`src/analytics/events.ts` — exhaustive catalog.** Correct and the most important one. `readonly DomainEventName[]` only checks each entry *is* a valid name, never that all are present, so a new `StratumEvent` member could ship, be exported by `captureDomainEvent` (which derives the name from `event.type` at runtime), and never reach the catalog or the FAQ — defeating the whole point of the docs-drift test. My docblock claimed the opposite. Now `satisfies Record<DomainEventName, true>`; I verified it by deleting a key and confirming `tsc` fails with `TS1360`.

- **`src/analytics/tracker.ts` — id-less non-system actor.** Correct, and I checked the claim: `src/services/change-flow.ts:334` and `:586` build `{ type: "user" }` with the id omitted when there is none. The `|| !event.actorId` branch routed that to the non-opted-out system actor, so a real person's activity was exported with no preference consulted — the opt-out defeated by a missing field rather than a decision. Now fails closed like every other unresolvable lookup, with a regression test.

- **`src/analytics/tracker.ts` — construction boundary.** Correct in the sense that matters: the class docblock claims a caller "cannot reach `capture()` without having resolved the preference", but a public `static create` accepted any caller-built actor. The class *value* is no longer exported (only its type, which `webhook-delivery` and the MCP route need), so construction is module-private and compiler-enforced. It promptly rejected two of my own tests, which now go through the real factories.

- **`src/mcp/outcome.ts` — unbounded `mcp_method`.** Correct. `message.method` is caller-supplied; any authenticated client can POST an arbitrary string, get `METHOD_NOT_FOUND`, and have that string land in the payload — exactly what the catalog's "source-code literal or bounded value" rule forbids. Now bounded to the methods `handleMessage` implements, anything else `unknown`. FAQ updated to say so.

- **`tests/analytics-catalog.test.ts` — assertion with no matcher.** Correct, my sloppiness: `expect(...)` with nothing after it asserts nothing.

**Declined, with reasons**

- **Restricting `createPostHogClient`.** It has to stay exported — `tracker.ts` imports it from `posthog.ts`. The construction path that actually mattered is closed above, and the module docblock already directs callers away from it.

- **`src/index.ts` — failing closed on an unresolved `telemetryOptOut`.** I traced every assignment: no path sets `userId` without also setting `telemetryOptOut`. The session branch in particular returns 401 *before* setting either when the user lookup fails or throws (`src/middleware/auth.ts:432-446`), and the token and OAuth branches set both together. So the only unresolved case is an unauthenticated caller, who is already captured personless under the `server` sentinel with no person data attached. Suppressing there would drop `error_occurred` for every anonymous 5xx — the precise signal that event was added to capture, since an unhandled exception never reaches the analytics middleware's post-`next()` body at all.

Gates: typecheck, lint, 3551 unit tests pass. (The 43 failures in `tests/issues*.test.ts` are pre-existing — they fail identically on the base commit and are untouched here.)

jlamoreaux:
@coderabbit review

jlamoreaux:
> 
> 
> <details>
> <summary>⚠️ Action not completed</summary>
> 
> Review rate limited.
> 
> > Note: CodeRabbit is an incremental review system and does not re-review already reviewed commits. This command is applicable only when automatic reviews are paused.
> 
> </details>

@coderabbit i didn't think you reviewed this commit yet?

jlamoreaux:
@coderabbit approve if nothing else is blocking

jlamoreaux:
@coderabbit approve

